PR #17, opened 2018-07-02
change Host label to Node label and change host api to node api

ly5156:
[96](https://github.com/rancher/longhorn/issues/96)


loganhz:
LGTM

PR #24, opened 2018-07-09
Add one additional column Robustness for the volume list and volume detail page

ly5156:
https://github.com/rancher/longhorn/issues/104#issuecomment-403182198


loganhz:
LGTM

PR #27, opened 2018-07-10
Consume websocket apis

LLParse:
Issues:
* Inbound websocket changes cause user input to be erased. This is particularly annoying for modal popups which close/reopen, losing all state.
* roadhog dev server doesn't forward websockets (need layer 4 forwarding), workaround should be to ignore CORS and create sockets directly to LONGHORN_MANAGER_IP - currently hardcoded to 127.0.0.1

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 src/**/*-test.js
+src/utils/websocket.js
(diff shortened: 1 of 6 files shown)


loganhz:
@ly5156 will fix the first issue.
I will submit another PR to fix the second issue.

And we need to merge the backend change first.
Thanks

yasker:
Merged, thanks.

PR #41, opened 2018-07-24
Scheduling failure for volume only when status is False

ly5156:
https://github.com/rancher/longhorn/issues/120


yasker:
Merged, thanks.

PR #48, opened 2018-07-27
Fix storageReserve minimal

yasker:
Should be 0


yasker:
https://github.com/rancher/longhorn/issues/124

PR #54, opened 2018-08-03
Page tweaks

ly5156:
https://github.com/rancher/longhorn/issues/157


yasker:
Health unknown is still showing for the volume details page.

PR #57, opened 2018-08-04
websocket - clientside reconnect logic

ly5156:
https://github.com/rancher/longhorn/issues/145

### DIFF
--- a/src/utils/websocket.js
+++ b/src/utils/websocket.js
@@ -1,3 +1,5 @@
+import ReconnectingWebSocket from 'reconnecting-websocket'
+
 function constructWebsocketURL(type, period) {
   let loc = window.location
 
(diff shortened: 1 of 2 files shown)


src/utils/websocket.js
```diff
@@ -13,11 +15,15 @@ function constructWebsocketURL(type, period) {
 
 export function wsChanges(dispatch, type, period) {
   let url = constructWebsocketURL(type, period)
-  const ws = new WebSocket(url)
-  ws.onmessage = function (msg) {
+  const options = {
+    connectionTimeout: 10000,
+    maxRetries: 10,
```
LLParse: Can we remove this line? The default is `Infinity` with geometric backoff, which is what we want.

This is my test:
1. Null route websocket traffic (however you want to do it, deleting manger is fine too)
2. Refresh UI page
3. Wait 101 seconds
4. Restore websocket connectivity

yasker: Yeah, that sounds good.

@LLParse can you fork this PR and update it? Remember give credit to @ly5156 .

src/utils/websocket.js
```diff
@@ -13,11 +15,15 @@ function constructWebsocketURL(type, period) {
 
 export function wsChanges(dispatch, type, period) {
   let url = constructWebsocketURL(type, period)
-  const ws = new WebSocket(url)
-  ws.onmessage = function (msg) {
+  const options = {
+    connectionTimeout: 10000,
```
LLParse: The default of `4000` is more responsive. I've heard the figure `500ms` quoted as a reasonable inter-continental RTT. There could be other high latency situations that I'm not considering.

yasker:
Close this in favor of https://github.com/rancher/longhorn-ui/pull/69

PR #61, opened 2018-08-06
Live Volume icon doesn't look right

ly5156:


yasker:
Didn't see it in the test environment?

PR #62, opened 2018-08-06
filter out same directory when edit disks on node

ly5156:
https://github.com/rancher/longhorn/issues/170

### DIFF
--- a/src/routes/host/EditDisk.js
+++ b/src/routes/host/EditDisk.js
@@ -22,7 +22,7 @@ const modal = ({
     const isStorageChanged = (origValue, inputValue) => {
       return formatMib(origValue) !== formatMib(inputValue)
     }
-    validateFields((errors, values) => {
+    validateFields({ force: true }, (errors, values) => {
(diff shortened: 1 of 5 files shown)


yasker:
1. If the input directory doesn't match the existing directory with the last `/` (e.g. `/var/lib/rancher/longhorn` is the same as `/var/lib/rancher/longhorn/`), input validation doesn't work. UI should add `/` in the end automatically if it doesn't exist.
2. If input the same path, the warning shows. But if click the save now regardless, the warning shows up in the existing disk as well. Then remove the new disk. The warning for the existing disk is still there. Still cannot save.

yasker:
`/var/lib/rancher/longhorn///////` is the same as `/var/lib/rancher/longhorn/`.

In fact, we should remove all the `/` at the end of the path for validation. But we should send the disk path to the backend with single `/` at the end of the path (to indicate it's a directory), no matter how many (or none) `/` at the end of the user inputted string.

PR #63, opened 2018-08-06
Enhance the dashboard

ly5156:
- [x] All volumes status

- [x] All nodes storage status

- [x] All nodes status

- [x] Click the items below or the part of graphic, jump to the filtered list of the volume or node page

https://github.com/rancher/longhorn/issues/122


yasker:
- [ ] When there is no volume/node or one, use the singular form `volume/node`, otherwise use the plural form `volumes/nodes`.
- [ ] Keep a consistent order for the different colors. e.g. Green/Yellow/Blue/Grey/Red (the order may change later because of the design)
- [ ] The calculation I gave you for availableStorage and used storage is wrong. It should be:
```
AvailableForSchedulingStorage = sum(enabledNodes.enabledDisks.AvailableStorage - enabledNodes.enabledDisks.ReservedStorage)
UsedStorage = sum(enabledNodes.enabledDisk.MaximumStorage - enabledNodes.enabledDisks.AvailableStorage)
```

yasker:
- [ ] Volume Status:
```
Volume Status
    1. Total. Total number of the volumes
    2. Healthy (green)
        a. Volume.State == attached && volume.Robustness == Healthy
    3. Degraded (yellow)
        a. Volume.State == attached && volume.Robustness == Degraded
    4. In Progress (blue)
        a. (Volume.State != attached and volume.State != detached) or (volume.State == attached && volume.Robustness != Healthy && volume.Robustness != Degraded)
    5. Detached (grey)
        a. Volume.State == detached && volume.Robustness != Fault
    6. Fault (red)
        a. Volume.State == detached && volume.Robustness == Fault
```

yasker:
- [ ] When there is no volume, shows `No Volume` or `0 volume`, instead of `No Data`.
- [ ] Seems I cannot trigger `Unscheduable`? Set reserved space more than available space, it's still doesn't show in yellow. Maybe need to check with Jacie.
- [ ] Change Storage `Available` to `Schedulable`. The name `Schedulable` is more more appropriate, because it's `disk.Available - disk.Reserved`.
- [ ] Rename `Scheduling Disabled` to `Disabled`.
- [ ] Storage dial cannot be clicked. So don't show `hand` icon.
- [ ] Volume node filter doesn't work.

yasker:
- [ ] Put schedulable in the node page to the leftmost column instead of the rightmost column.
- [ ] Add disk status.

yasker:
Filed the remaining issues into separate issues.

PR #64, opened 2018-08-06
Create info modal when create backup is clicked

LLParse:
https://i.imgur.com/F5TZoPg.png


yasker:
Does this happen before the backup API call completed or after? It should be after.

Also, can we add this modal to the backup from clicking `snapshot->backup` as well?

PR #74, opened 2018-08-09
Replica should be associated with the disk as well

ly5156:
https://github.com/rancher/longhorn/issues/188


yasker:
- [ ] Click the right upper corner `x` to close the disk edit modal result in the page blurry.

yasker:
I will merge this, and we can fix the blurry issue in another PR.

PR #79, opened 2018-08-13
New node page design

ly5156:
https://github.com/rancher/longhorn/issues/214


yasker:
- [ ] The `Disks` subtitle and the shadow was missing
- [ ] The bar seems thinner than the design
- [ ] Sorting not available for many fields.
- [ ] The second line of text should be similar than the first line.
- [ ] Only node level has the yellow indicator when almost full

Also:
- [ ] We can add the `Expand All` button on the top left following this style (not the blue frame, that's selected)
image 2

PR #80, opened 2018-08-13
Don't allow attach to the unready node

ly5156:
https://github.com/rancher/longhorn/issues/205


yasker:
@JacieChao Can you help to test this?

JacieChao:
@yasker Tested, UI could filter out `unready` node when attach volume.

PR #82, opened 2018-08-15
Add event log to the dashboard

ly5156:
https://github.com/rancher/longhorn/issues/186
Tested with backend https://github.com/rancher/longhorn-manager/pull/172


ly5156:
@yasker

PR #83, opened 2018-08-15
Form validation

ly5156:
1. volume number of replica limit 2 to 10 and must be a positive integer
2. The volume size must be greater than 0


yasker:
- [ ] Don't autocorrect the number of replicas. Just remind the user and invalid the input.

PR #86, opened 2018-08-20
Volume page improvement

ly5156:
1. sort by volume created time default in volume page 
2. Add volume size and replicas validation 
3. change 'create backup' button icon to copy icon in volume detail page
4. Do not clear selected item when hide salvage modal and the text of replica item overflow(https://github.com/rancher/longhorn/issues/234)


ly5156:
@yasker

PR #87, opened 2018-08-21
User experience in the recurring snapshot/backup page

ly5156:
https://github.com/rancher/longhorn/issues/236


ly5156:
@yasker

ly5156:
@yasker This scene is similar to the edit disks in node page, only edit row has a cancel button, and new row only has a remove button.
Or add a cancel button to left of save button?

yasker:
We need to keep it consistent styles with the node disk edit page.

For example, node disk edit page has one action to enable the edit for the whole page. We don't need to do pop out here, but one action to start editing the recurring list would be helpful.

Also, there are cancel/save options.

The deleted line wasn't obvious from UI perspective. 

I think we can just copy what we've done with the disks editing pop out page and put it here?

PR #99, opened 2018-10-26
Improve Node UX

LLParse:
https://github.com/rancher/longhorn/issues/306

### DIFF
--- a/src/assets/styles/antd-customize/generic.less
+++ b/src/assets/styles/antd-customize/generic.less
@@ -26,7 +26,7 @@
 
     &.ant-table-small {
       .ant-table-thead > tr > th {
-        background: #f7f7f7;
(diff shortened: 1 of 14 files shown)


LLParse:
@yasker Ready for review

yasker:
Either don't show `Remove Node` when it's not possible to remove or add a tooltip say something like `The node must be removed from Kubernetes first` when users hover over the action. Otherwise, users would be confused about why this action is not available.

LLParse:
@yasker Ready for review

yasker:
The `delete` option in the dropdown for a replica in volume attaching/detaching state should also be disabled. And we can add a tooltip there as well.

Other than that, LGTM.

LLParse:
@yasker

PR #101, opened 2018-11-08
Add upgrade icon to footer when new version is available

LLParse:
https://github.com/rancher/longhorn/issues/312


yasker:
Can we do a flat icon? There is no other 3d icon in the UI.

LLParse:
@yasker Ready for review

PR #108, opened 2019-01-11
volume-head should be reported as a part of snapshot list

ly5156:
https://github.com/rancher/longhorn/issues/364


ly5156:
@yasker

yasker:
Close in favor of #109

PR #112, opened 2019-02-13
add image upgradable and replica redundancy tooltip in volume page

ly5156:
https://github.com/rancher/longhorn/issues/383
https://github.com/rancher/longhorn/issues/382


yasker:
Looks really good!

A few small issues.
- [ ] If I click filter and select e.g. `Node redundancy: no`, then click `Go`, then click `Volume`, the input of `no` would remain in the text.
- [ ] Change the order of dropdown for `Node redundancy` from `Yes`, `No`, `Limited` to `Yes`, `Limited`, `No`.
- [ ] I realized that since the health of the volume can be unknown (blank), it looks better to put those the status before the health instead of after, e.g. "<icon>Healthy`.
- [ ] Also I realized that engine upgrade doesn't relate to the health or not of the volume. How about we put the engine upgrade icon in front of the `State` instead?

yasker:
Merged, thanks!

PR #129, opened 2019-04-16
Fixed the workload/pod field 

smallteeths:
https://github.com/rancher/longhorn/issues/461

### DIFF
--- a/src/assets/styles/antd-customize/tree.less
+++ b/src/assets/styles/antd-customize/tree.less
@@ -50,7 +50,6 @@
     .tree-snapshot-start-point{
       width: 15px;
       height: 15px;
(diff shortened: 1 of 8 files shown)


yasker:
- [ ] In the overview page, it's unclear to users, that `workload/pod` entry can be clicked and get more details. Users may think it would click to the volume page. It's better if we can add a tooltip for the entry. If the entry is historical, said: `Last time used: xxx. Click for details`; if not, just say `Click for details`. Add a 0.5 second delay to the tooltip as well.
- [ ] In the detail page, there is a group for workload status, but it's a bit hard to observe. Maybe make it slightly more obvious?

PR #130, opened 2019-04-17
add recurring snapshot and backups to the volumelist

smallteeths:
https://github.com/rancher/longhorn/issues/83#issuecomment-480435816


yasker:
- [ ] Put the column after `Created`, as the last column in the list
- [ ] The column name can be `Recurring Schedule`. And don't separate a word into two lines.
- [ ] The icon is nice. But a backup schedule is different from a snapshot schedule. So if it only contains snapshot schedules (less safe), use yellow color, and said `Only scheduled snapshot` in the tooltip; if it contains at least one backup schedule, use blue color.
- [ ] For the pop-out window: 1. don't need to show the name, since we don't provide that information in the volume detail page to the user; 2. Put the type as the first entry, to keep consistent with the volume detail page. And we can add the same icon (blue for backup, yellow for snapshot) at the end of the field, e.g. `Type: Backup <BLUE ICON>`. 3. If there are multiple snapshots mixed with backups, show backups first. Snapshots can be further to the right.
- [ ] `Last Backup At` field wasn't shown in the Volume list page. Both `Last Backup` and `Last Backup At` wasn't shown in the detail page (use `Last Backup` instead of `Latest Backup`)

PR #131, opened 2019-04-18
Add fields to the engine list

smallteeths:
add Reference Count,Last Reference At to engine image list


yasker:
- [ ] The field name `Reference Count At` is wrong. It should be `Last Reference at` or `No Reference Since`. But thinking about it, let's remove this field. It would be good for the user to know when is the last time engine image was used, but the timestamp here can be confusing.

PR #140, opened 2019-05-08
Merge the State with Health. Name the new field State

smallteeths:
https://github.com/rancher/longhorn/issues/510


yasker:
- [ ] Keep `on` as the prefix for the attached node, even the volume doesn't have associated pod. We need to make sure users know it's a node instead of a pod.

PR #144, opened 2019-05-12
Create Disaster Recovery Volume

smallteeths:
https://github.com/rancher/longhorn/issues/495

### DIFF
--- a/src/components/Icon/IconStandBackup.js
+++ b/src/components/Icon/IconStandBackup.js
@@ -1,0 +1,18 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+function IconBackup({ width = 18, height = 18, fill = '#00C1DE' }) {
+  return (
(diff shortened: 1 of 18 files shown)


yasker:
-[ ] BackupVolume overview page: Don't show backup URL in the dialog. Base image and size should be filled automatically using the information from the backup volume, and they cannot be modified. (Notice the size of volume can be less than 1 Gi)
-[ ] BackupVolume overview page: Number of replicas in the dialog should follow the setting `Default Replica Count`.
-[ ] Volume overview page: Action menu:  Allow show `Backups` for the DR volume. `To normal volume` should be named `Activate Disaster Recovery Volume`, and don't show the option for the normal volume (since they won't able to use the option in any situation)
-[ ] Volume overview page: Change the icon for DR volume to something like downloading from the cloud, e.g. https://www.flaticon.com/free-icon/cloud-download-symbol_33651#term=download%20cloud&page=1&position=68

PR #150, opened 2019-05-15
Add restoring process state to the volume list

smallteeths:
https://github.com/rancher/longhorn/issues/534


yasker:
@shuo-wu verified it works.

PR #164, opened 2019-06-06
Add able to filter volumes by namespace in the volumes list

smallteeths:
https://github.com/rancher/longhorn/issues/577


yasker:
- [ ] If the filter dropdown started with `Name`, the field choices with more than 4 characters cannot be seen in full in the dropdown. (Change can be added as another commit for this PR.)

PR #167, opened 2019-06-19
Remove roadhog plugin

smallteeths:
Roadhog into webpack


yasker:
There are some Chinese comments in the PR? Is it from upstream or we added it?

PR #172, opened 2019-06-27
support for storage pools

smallteeths:
https://github.com/longhorn/longhorn/issues/311


yasker:
- [ ] `Edit disks` changes to `Edit node and disks`
- [ ] Move tags in the `Edit node and disks` to below the path.
- [ ] `Configure Node Schedudling` to `Node Scheduling`
- [ ] `Configure Node Tags` to `Node Tags`
- [ ] Show node and disk tags in the node overview.

PR #179, opened 2019-07-25
Add action to delete all the backups of one volume

smallteeths:
https://github.com/longhorn/longhorn/issues/598


yasker:
-[ ] Text `Delete all backups of the volume`, for the pop-out window.

PR #180, opened 2019-07-29
Disable base image feature for now

smallteeths:
https://github.com/longhorn/longhorn/issues/650


yasker:
- [ ] Need to remove the option. We're not sure when we reintroduced the feature it would use the same option.

PR #182, opened 2019-08-01
Add progress for restore

smallteeths:
https://github.com/longhorn/longhorn/issues/49


yasker:
- [ ] If restoration is in progress, don't allow snapshot operations and recurring backup setting, etc.

src/routes/volume/VolumeActions.js
```diff
@@ -71,14 +71,25 @@ function actions({ selected, engineImages, showAttachHost, detach, showEngineUpg
     }
   }
 
+  const isRestoreIng = () => {
```
yasker: `restoring`

PR #190, opened 2019-08-05
Allow user to disable the scheduling when the node is down

smallteeths:
Allow user to disable the scheduling when the node is down


yasker:
@shreesha21 ^^

PR #192, opened 2019-08-06
Stop websocket updates when data is not up-to-date

smallteeths:
https://github.com/longhorn/longhorn/issues/648


yasker:
You meant `drop` I think? It only applies to one update right?

Any side effect observed?

smallteeths:
The prerequisite for fixing this bug is , As long as any piece of data updated, data returned timestamp of any piece of data should be up to date.I compared the data and it was true and
No side effect observed @yasker

PR #211, opened 2019-11-05
fix(take snapshot) Remove the field MigrationNodeID

smallteeths:
https://github.com/longhorn/longhorn/issues/848


yasker:
Seems GitHub bugged out. The commit has been merged.

PR #217, opened 2019-11-25
Fix(snapshot) Fix the snapshot reverting message

smallteeths:
https://github.com/longhorn/longhorn/issues/897


yasker:
- [ ] There is no need for `Cancel` and `OK` when reverting is not allowed. Just put an `OK` there.
- [ ] `Attached Node` is not shown for volume in maintenance mode.

PR #227, opened 2019-12-26
feat(key down) Hold CTRL to bypass confirmation

smallteeths:
https://github.com/longhorn/longhorn/issues/944
https://github.com/longhorn/longhorn/issues/942


meldafrawi:
This PR support Hold CTRL to bypass confirmation except for Volume Details page.
Support for Volume Details page will be addressed in another PR.

PR #246, opened 2020-03-16
Fix(snapshot) Add json format judgments

smallteeths:
https://github.com/longhorn/longhorn/issues/1082


yasker:
@smallteeths Need `signed-off-by`. Otherwise LGTM since the user verified the fix at https://github.com/longhorn/longhorn/issues/1082#issuecomment-601714220 .

PR #248, opened 2020-03-31
feat(Disks) Changed DiskUpdate Api parameter type from Array to Object

smallteeths:
https://github.com/longhorn/longhorn/issues/972


yasker:
- [ ] For the new disk added by the user, we can allow the user to customize the disk name, just make sure it fits the same standard as the volume name.
- [ ] `disk-NaN` is a weird default name.

PR #253, opened 2020-04-17
fix(node, volume-detail): fix text error and add tooltip for rebuilding replica

ly5156:
fix(node, volume-detail): fix text error and add tooltip for rebuilding replica

Linked Issues:
https://github.com/longhorn/longhorn/issues/1167
https://github.com/longhorn/longhorn/issues/1204  issue-#3


yasker:
Can you merged the tooltip and the progress bar for rebuilding?

PR #262, opened 2020-05-08
Show volume related events in the volume details

smallteeths:
longhorn/longhorn#990

image: wangsiye/longhorn-ui:c45d726


yasker:
- [ ] Can we make clicking the `+` to expand, instead of clicking the `Expand` button. It's consistent with our node page.

PR #265, opened 2020-05-11
Fix issues 1290 1211 1201

smallteeths:
1. Fix node is still schedulable When there are disks with disabled scheduling and invalid disks
2. Resotre last backup is disabled when an error occurs in the backup file
3. Fixed the field style of Edit Node and Disks popup

https://github.com/longhorn/longhorn/issues/1201
https://github.com/longhorn/longhorn/issues/1211
https://github.com/longhorn/longhorn/issues/1290

image: wangsiye/longhorn-ui:187f116


yasker:
Regarding #265 (saw a screenshot only)
- [ ] Put the Storage Available at the last line instead of the first one.

PR #267, opened 2020-05-13
Show error and disable restore when there is Invalid backup file in S3

smallteeths:

image: wangsiye/longhorn-ui:79c09d9

https://github.com/longhorn/longhorn/issues/1321
https://github.com/longhorn/longhorn/issues/1320


yasker:
Verified:
- [ ] longhorn/longhorn#1321
- [ ] longhorn/longhorn#1320

PR #268, opened 2020-05-14
feat(volume state) Add field ready showing whether the volume is read…

smallteeths:
…y for workload

image: wangsiye/longhorn-ui:8de351f

https://github.com/longhorn/longhorn/issues/1260


yasker:
- [ ] Update `NO` or `YES` in the volume detail page to `Ready` and `Not Ready`

PR #272, opened 2020-05-19
feat(conditions) Show up the volume, node, disk conditions

smallteeths:
https://github.com/longhorn/longhorn/issues/1359

image: wangsiye/longorn-ui:3f5515c

### DIFF
--- a/src/models/snapshot.js
+++ b/src/models/snapshot.js
@@ -164,8 +164,10 @@ export default (namespace) => {
         payload,
       }, { put }) {
         const data = payload.volume
-        yield put({ type: 'setVolume', payload: data })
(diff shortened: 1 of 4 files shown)


yasker:
Here are the current conditions we're using for:

### Node ###
1. Ready: Green for true, red for false.
2. Mount Propagation: Green for true, red for false.
3. Schedulable: Green for true, red for false.

### Disk ###
1. Ready: Green for true, red for false.
2. Scheduable: Green for true, red for false.

### Volume ###
1. Scheduled: Green for true, red for false.
2. Restore: Green for true, **grey for false if Reason is empty (no restore needed)**, red for false if the Reason is not empty.

yasker:
LGTM.

PR #274, opened 2020-05-20
feat(settings) Use the dropdown component for setting item 'VolumeAtt…

smallteeths:
https://github.com/longhorn/longhorn/issues/1368

image: wangsiye/longorn-ui:3f5515c

### DIFF
--- a/src/routes/setting/setting.js
+++ b/src/routes/setting/setting.js
@@ -1,9 +1,10 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { Form, Input, Button, Spin, Icon, Checkbox } from 'antd'
+import { Form, Input, Button, Spin, Icon, Checkbox, Select } from 'antd'
 import styles from './setting.less'
 import { classnames } from '../../utils'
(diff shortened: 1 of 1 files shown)


yasker:
LGTM.

src/routes/setting/setting.js
```diff
@@ -28,6 +29,13 @@ const form = ({
     return rules
   }
   const genInputItem = (setting) => {
+    if (setting.id === 'volume-attachment-recovery-policy') {
```
joshimoo, line 32: Instead of hardcoding this behavior for the volume-attachment-recovery-policy id, consider checking setting.definition.options. Than you can have a generic dropdown and the frontend is not dependent on the backend id or the option values.

Pseudo Code below
```
if (setting.definition.options) {
return 
(<Select>
        for each option : setting.definition.options {
                <Option value="$option">$option</Option>
        }
</Select>)
}
```

yasker: Good catch @joshimoo 

@smallteeths can we do this instead?

PR #276, opened 2020-05-21
fix(replicas) Fix Volume status tooltip error when Replica Node Level…

smallteeths:
… hard anti affinity is set

image: wangsiye/longhorn-ui:ff585c6

https://github.com/longhorn/longhorn/issues/1364


yasker:
Signed-off check failed.

PR #290, opened 2020-06-04
fix(build path) Fix absolute path to relative path

smallteeths:


yasker:
Can you explain a bit more about this PR?

PR #297, opened 2020-06-28
fix(Dockerfile) Reduce image size

smallteeths:

https://github.com/longhorn/longhorn/issues/1417

image: wangsiye/longhorn-ui:0bc7973


yasker:
Did we keep basic network troubleshooting command e.g. curl/ping in the image?

PR #298, opened 2020-07-02
Fix(antd) Upgrade antd verison 3.19.4 to 3.26.16

smallteeths:

image:  wangsiye/longhorn-ui:cac7544

https://github.com/longhorn/longhorn/issues/1329


yasker:
What's the impact? Security issue fix?

PR #301, opened 2020-07-09
feat(volume) Change the prompt text of the delete operation according…

smallteeths:
image: wangsiye/longhorn-ui:bd573fe

https://github.com/longhorn/longhorn/issues/520

### DIFF
--- a/src/routes/volume/VolumeActions.js
+++ b/src/routes/volume/VolumeActions.js
@@ -5,6 +5,41 @@ import { DropOption } from '../../components'
 const confirm = Modal.confirm
 
 function actions({ selected, engineImages, showAttachHost, detach, showEngineUpgrade, deleteVolume, showBackups, showSalvage, rollback, showUpdateReplicaCount, showExpansionVolumeSizeModal, showCancelExpansionModal, createPVAndPVC, changeVolume, confirmDetachWithWorkload, commandKeyDown }) {
+  const deleteWranElement = (record) => {
+    let workloadResources = ''
+    let hasPvTooltipText = ''
+    let pvResources = ''
+
+    if (record && record.kubernetesStatus && record.kubernetesStatus.pvStatus && record.kubernetesStatus.pvName) {
+      hasPvTooltipText = (<div>{`The following resources associated with this volume (${record.name}) will be deleted:`}</div>)
+      pvResources = (<div style={{ margin: '10px', color: '#b7b7b7' }}>
+        <div style={{ paddingLeft: '10px' }}>PV Name: {record.kubernetesStatus.pvName}</div>
+        { !record.kubernetesStatus.lastPVCRefAt && record.kubernetesStatus.pvcName ? <div style={{ paddingLeft: '10px' }}>PVC Name: {record.kubernetesStatus.pvcName}</div> : ''}
+      </div>)
+    }
+    if (record.kubernetesStatus && record.kubernetesStatus.workloadsStatus && record.kubernetesStatus.workloadsStatus.length > 0) {
+      workloadResources = (<div>
+        <div style={{ margin: '10px 0px', color: '#b7b7b7' }}>{record.kubernetesStatus.workloadsStatus.map((item, i) => {
+          return (<div key={i} style={{ paddingLeft: '10px' }}>Pod Name: {item.podName} ({item.podStatus})</div>)
+        })}</div>
+      </div>)
+    }
+    if (record.state === 'attached') {
+      return (<div>
+          <div>{`The volume is attached on ${record.controllers && record.controllers[0] && record.controllers[0].hostId ? record.controllers[0].hostId : ''}. Deleting this volume may cause errors for any running applications using this volume!`} </div>
+          {workloadResources}
+          {hasPvTooltipText}
+          {pvResources}
+          <div style={{ marginTop: 10 }}>{`Are you sure you want to delete volume (${record.name}) ?`} </div>
+        </div>)
+    } else {
+      return (<div>
+        {hasPvTooltipText}
+        {pvResources}
+        <div style={{ marginTop: hasPvTooltipText || pvResources ? 10 : 0 }}>{`Are you sure you want to delete volume ${record.name} ?`} </div>
+      </div>)
+    }
+  }
   const handleMenuClick = (event, record) => {
     switch (event.key) {
       case 'attach':
@@ -17,8 +52,10 @@ function actions({ selected, engineImages, showAttachHost, detach, showEngineUpg
         if (commandKeyDown) {
           deleteVolume(record)
         } else {
+          let title = deleteWranElement(record)
           confirm({
-            title: `Are you sure you want to delete volume ${record.name} ?`,
+            width: 700,
+            title,
             onOk() {
               deleteVolume(record)
             },
--- a/src/routes/volume/VolumeBulkActions.js
+++ b/src/routes/volume/VolumeBulkActions.js
@@ -6,14 +6,48 @@ import style from './VolumeBulkActions.less'
 const confirm = Modal.confirm
 
 function bulkActions({ selectedRows, engineImages, bulkDeleteVolume, showBulkEngineUpgrade, showBulkChangeVolume, showBulkAttachHost, bulkDetach, bulkBackup, bulkExpandVolume, createPVAndPVC, createSchedule, confirmDetachWithWorkload, commandKeyDown }) {
+  const deleteWranElement = (rows) => {
+    let workloadResources = []
+    let pvResources = []
+    let hasVolumeAttached = false
+
+    rows.forEach((item) => {
+      if (item && item.kubernetesStatus && item.kubernetesStatus.pvStatus && item.kubernetesStatus.pvName) {
+        pvResources.push((<div style={{ margin: '10px', color: '#b7b7b7' }}>
+          <div style={{ paddingLeft: '10px' }}>PV Name: {item.kubernetesStatus.pvName}</div>
+          { !item.kubernetesStatus.lastPVCRefAt && item.kubernetesStatus.pvcName ? <div style={{ paddingLeft: '10px' }}>PVC Name: {item.kubernetesStatus.pvcName}</div> : ''}
+        </div>))
+      }
+      if (item.kubernetesStatus && item.kubernetesStatus.workloadsStatus && item.kubernetesStatus.workloadsStatus.length > 0) {
+        workloadResources.push((<div>
+          <div style={{ margin: '10px 0px', color: '#b7b7b7' }}>{item.kubernetesStatus.workloadsStatus.map((ele, i) => {
+            return (<div key={i} style={{ paddingLeft: '10px' }}>Pod Name: {ele.podName} ({ele.podStatus})</div>)
+          })}</div>
+        </div>))
+      }
+      if (item.state === 'attached') {
+        hasVolumeAttached = true
+      }
+    })
+    return (<div>
+      {hasVolumeAttached ? <div>Some of the selected volumes are attached to a node and may cause errors in any applications using them once they are deleted.</div> : ''}
+      {!hasVolumeAttached && workloadResources.length > 0 ? <div>The following applications used these volumes:</div> : '' }
+      {workloadResources.map((item, i) => <div key={i}>{item}</div>)}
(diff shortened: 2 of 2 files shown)


src/routes/volume/VolumeActions.js
```diff
@@ -5,6 +5,42 @@ import { DropOption } from '../../components'
 const confirm = Modal.confirm
 
 function actions({ selected, engineImages, showAttachHost, detach, showEngineUpgrade, deleteVolume, showBackups, showSalvage, rollback, showUpdateReplicaCount, showExpansionVolumeSizeModal, showCancelExpansionModal, createPVAndPVC, changeVolume, confirmDetachWithWorkload, commandKeyDown }) {
+  const deleteWranElement = (record) => {
+    let workloadStatus = ''
+    let hasPvTootipText = ''
+    let pvStatus = ''
+
+    if (record && record.kubernetesStatus && record.kubernetesStatus.pvStatus) {
+      hasPvTootipText = (<div>{`The longhorn-manager will delete the Kubernetes resources associated with the Volume ${record.name}`}</div>)
```
ttpcodes, line 14: I think this can be rephrased to the following:
```
The following resources associated with this volume will be deleted:
...
```

src/routes/volume/VolumeActions.js
```diff
@@ -5,6 +5,42 @@ import { DropOption } from '../../components'
...
+      pvStatus = (<div style={{ marginTop: '10px', color: '#b7b7b7' }}>
+        <div style={{ paddingLeft: '10px' }}>PV Name: {record.kubernetesStatus.pvName}</div>
+        <div style={{ paddingLeft: '10px' }}>PVC Name: {record.kubernetesStatus.pvcName}</div>
+        <div style={{ paddingLeft: '10px' }}>PV Status: {record.kubernetesStatus.pvStatus}</div>
```
ttpcodes, line 18: The `PV Status` shouldn't be needed I think since it's not an actual resource that gets deleted.

src/routes/volume/VolumeActions.js
```diff
@@ -5,6 +5,42 @@ import { DropOption } from '../../components'
...
+    if (record.kubernetesStatus && record.kubernetesStatus.workloadsStatus && record.kubernetesStatus.workloadsStatus.length > 0) {
+      workloadStatus = (<div>
+        <div style={{ marginTop: '10px', color: '#b7b7b7' }}>{record.kubernetesStatus.workloadsStatus.map((item, i) => {
+          return (<div key={i} style={{ paddingLeft: '10px' }}>Pod Name: {item.podName} ({item.podStatus})</div>)
+        })}</div>
+      </div>)
+    }
+    if (record.state === 'attached') {
+      return (<div>
+          <div>{`The Volume is Attached on ${record.controllers && record.controllers[0] && record.controllers[0].hostId ? record.controllers[0].hostId : ''} and that deleting the Volume can lead to errors in any applications using the Volume`} </div>
```
ttpcodes, line 30: I think this can be rephrased to:
```
The volume is attached on ... Deleting this volume may cause errors for any running applications using this volume!
```

src/routes/volume/VolumeActions.js
```diff
@@ -5,6 +5,42 @@ import { DropOption } from '../../components'
...
+          {hasPvTootipText}
```
ttpcodes, line 31: Should be reordered:
```
{workloadStatus}
{hasPvTooltipText}
{pvStatus}
```

src/routes/volume/VolumeBulkActions.js
```diff
@@ -13,7 +13,8 @@ function bulkActions({ selectedRows, engineImages, bulkDeleteVolume, showBulkEng
           bulkDeleteVolume(selectedRows)
         } else {
           confirm({
-            title: `Are you sure you want to delete volume(s) ${selectedRows.map(item => item.name).join(', ')} ?`,
+            width: 700,
+            title: `The longhorn-manager will delete the Kubernetes resources associated with the Volume ${selectedRows.map(item => item.name).join(', ')},Are you sure you want to delete volume(s) ${selectedRows.map(item => item.name).join(', ')} ?`,
```
ttpcodes, line 17: Is there a way to check if any of the to be deleted `Volumes` are attached or have resources associated with them and to only append that to the error message if that's the case like for single `Volumes`?

If not, I think this should be rewritten to be:
```
Any resources (PersistentVolumes and PersistentVolumeClaims) associated with these volumes will be deleted. Any volumes currently attached to a node may cause errors in any applications using them once they are deleted! Are you sure you want to delete volume(s) ... ?
```

src/routes/volume/VolumeBulkActions.js
```diff
@@ -6,14 +6,43 @@ import style from './VolumeBulkActions.less'
...
+      if (item.kubernetesStatus && item.kubernetesStatus.workloadsStatus && item.kubernetesStatus.workloadsStatus.length > 0) {
+        workloadResources.push((<div>
+          <div style={{ margin: '10px 0px', color: '#b7b7b7' }}>{item.kubernetesStatus.workloadsStatus.map((ele, i) => {
+            return (<div key={i} style={{ paddingLeft: '10px' }}>Pod Name: {ele.podName} ({ele.podStatus})</div>)
+          })}</div>
+        </div>))
+      }
+    })
+    return (<div>
+      <div>Any volumes currently attached to a node may cause errors in any applications using them once they are deleted!</div>
```
ttpcodes, line 29: Is this able to be conditionally displayed? (So that the message only displays if there are `Volumes` attached) If so, change the message to:
```
Some of the selected volumes are attached to a node and may cause any applications using them once they are deleted!
```
If not, the current message is fine.

Is it also possible to list the attached `Volumes` here?

src/routes/volume/VolumeActions.js
```diff
@@ -5,6 +5,41 @@ import { DropOption } from '../../components'
...
+  const deleteWranElement = (record) => {
+    let workloadResources = ''
+    let hasPvTooltipText = ''
+    let pvResources = ''
+
+    if (record && record.kubernetesStatus && record.kubernetesStatus.pvStatus) {
+      hasPvTooltipText = (<div>{`The following resources associated with this volume (${record.name}) will be deleted:`}</div>)
+      pvResources = (<div style={{ margin: '10px', color: '#b7b7b7' }}>
+        <div style={{ paddingLeft: '10px' }}>PV Name: {record.kubernetesStatus.pvName}</div>
+        <div style={{ paddingLeft: '10px' }}>PVC Name: {record.kubernetesStatus.pvcName}</div>
```
ttpcodes, line 17: Is it possible to only show `PVC Name` if there is a `PersistentVolumeClaim`?

`PVC Name` should also only be shown if `lastPVCRefAt` is empty.

src/routes/volume/VolumeBulkActions.js
```diff
@@ -6,14 +6,43 @@ import style from './VolumeBulkActions.less'
...
+      {workloadResources.map((item, i) => <div key={i}>{item}</div>)}
+      <div>Any resources (PersistentVolumes and PersistentVolumeClaims) associated with these volumes will be deleted.</div>
```
ttpcodes, line 31: Same deal here. If you can conditionally display the message (so it will only be displayed if there are resources that will be deleted), change it to:
```
The following resources (PersistentVolumes and PersistentVolumeClaims) associated with these volumes will be deleted:
```
If not, the current message is fine.

src/routes/volume/VolumeBulkActions.js
```diff
@@ -6,14 +6,43 @@ import style from './VolumeBulkActions.less'
...
 function bulkActions({ selectedRows, engineImages, bulkDeleteVolume, showBulkEngineUpgrade, showBulkChangeVolume, showBulkAttachHost, bulkDetach, bulkBackup, bulkExpandVolume, createPVAndPVC, createSchedule, confirmDetachWithWorkload, commandKeyDown }) {
+  const deleteWranElement = (rows) => {
+    let workloadResources = []
+    let pvResources = []
+
+    rows.forEach((item) => {
+      if (item && item.kubernetesStatus && item.kubernetesStatus.pvStatus) {
+        pvResources.push((<div style={{ margin: '10px', color: '#b7b7b7' }}>
+          <div style={{ paddingLeft: '10px' }}>PV Name: {item.kubernetesStatus.pvName}</div>
+          <div style={{ paddingLeft: '10px' }}>PVC Name: {item.kubernetesStatus.pvcName}</div>
```
ttpcodes, line 17: Is it possible to only show `PVC Name` if there is a `PersistentVolumeClaim`?

`PVC Name` should also only be shown if `lastPVCRefAt` is empty.

src/routes/volume/VolumeBulkActions.js
```diff
@@ -6,14 +6,47 @@ import style from './VolumeBulkActions.less'
...
+            return (<div key={i} style={{ paddingLeft: '10px' }}>Pod Name: {ele.podName} ({ele.podStatus})</div>)
+          })}</div>
+        </div>))
+      }
+      if (item.state === 'attached') {
+        hasVolumeAttached = true
+      }
+    })
+    return (<div>
+      {hasVolumeAttached ? <div>Some of the selected volumes are attached to a node and may cause any applications using them once they are deleted!</div> : <div>Any volumes currently attached to a node may cause errors in any applications using them once they are deleted!</div>}
```
ttpcodes, line 33: I made a typo here. This should be:
```
Some of the selected volumes are attached to a node and may cause errors in any applications using them once they are deleted.
```
The other message starting with `Any volumes...` shouldn't be needed, even if none of the `Volumes` are attached.

@yasker Would it be good to list off the `Volumes` here that are attached for the user to easily see?

src/routes/volume/VolumeActions.js
```diff
@@ -5,6 +5,41 @@ import { DropOption } from '../../components'
...
+    if (record.kubernetesStatus && record.kubernetesStatus.workloadsStatus && record.kubernetesStatus.workloadsStatus.length > 0) {
+      workloadResources = (<div>
+        <div style={{ margin: '10px 0px', color: '#b7b7b7' }}>{record.kubernetesStatus.workloadsStatus.map((item, i) => {
+          return (<div key={i} style={{ paddingLeft: '10px' }}>Pod Name: {item.podName} ({item.podStatus})</div>)
+        })}</div>
+      </div>)
+    }
+    if (record.state === 'attached') {
+      return (<div>
+          <div>{`The Volume is attached on ${record.controllers && record.controllers[0] && record.controllers[0].hostId ? record.controllers[0].hostId : ''}. Deleting this volume may cause errors for any running applications using this volume!`} </div>
```
ttpcodes, line 29: `volume`, not `Volume`.

src/routes/volume/VolumeBulkActions.js
smallteeths: When the volume is not attached, but there is a workload associated with it. without `with Any volumes...` the prompt message will be very strange

smallteeths: Prompt `The following applications using these volumes:` when the volume is not atteched and there is an associated workload. is this OK ? @ttpcodes

yasker:
https://github.com/longhorn/longhorn/issues/1447

src/routes/volume/VolumeBulkActions.js
ttpcodes: If a workload like a `Pod` is using the `Volume`, then the `Volume` should be attached I believe. Right now when we detect associated workloads we only look for `Pods` and not other resources that may be using the `Volume`, like `Stateful Sets` and `Deployments` where the `Volume` could be defined but not currently in use. So I think the message I suggested should be fine.

yasker: It's a good idea if we can list the volumes here.

yasker:
@ttpcodes can you file a separate issue regarding the review comment that hasn't been addressed? I want to include the functionality in v1.0.1-rc1 first.

ttpcodes:
I've filed longhorn/longhorn#1561 with the unaddressed review comment.

yasker:
Thanks @ttpcodes

src/routes/volume/VolumeBulkActions.js
smallteeths: Can I understand it this way, when the volume is not attached, we do not need to display the workload related information. If the workload information does not have a title, the style will look strange. like this 
Jietu20200711-080022
@ttpcodes

PR #303, opened 2020-07-11
Add support for arm64 (draft)

ivang:
Addresses partially longhorn/longhorn#6 by adding support for arm64 (armhf not supported).

Update of the base image was needed because `nginx:1.18.0` adds support for arm64.

This PR should be viewed as part of longhorn/longhorn-engine#514.


yasker:
Close this PR since the issue has been addressed when we upgrading the base image for the build.

PR #305, opened 2020-07-14
fix(Delete message) Optimized the prompt message for deleting volume

smallteeths:
https://github.com/longhorn/longhorn/issues/1561

image: wangsiye/longhorn-ui:5008222

### DIFF
--- a/src/routes/volume/VolumeActions.js
+++ b/src/routes/volume/VolumeActions.js
@@ -24,21 +24,15 @@ function actions({ selected, engineImages, showAttachHost, detach, showEngineUpg
         })}</div>
       </div>)
     }
-    if (record.state === 'attached') {
-      return (<div>
-          <div>{`The volume is attached on ${record.controllers && record.controllers[0] && record.controllers[0].hostId ? record.controllers[0].hostId : ''}. Deleting this volume may cause errors for any running applications using this volume!`} </div>
-          {workloadResources}
-          {hasPvTooltipText}
-          {pvResources}
-          <div style={{ marginTop: 10 }}>{`Are you sure you want to delete volume (${record.name}) ?`} </div>
-        </div>)
-    } else {
-      return (<div>
-        {hasPvTooltipText}
-        {pvResources}
-        <div style={{ marginTop: hasPvTooltipText || pvResources ? 10 : 0 }}>{`Are you sure you want to delete volume ${record.name} ?`} </div>
-      </div>)
-    }
+
+    return (<div>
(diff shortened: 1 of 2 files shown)


src/routes/volume/VolumeActions.js
```diff
@@ -24,21 +24,14 @@ function actions({ selected, engineImages, showAttachHost, detach, showEngineUpg
         })}</div>
       </div>)
     }
-    if (record.state === 'attached') {
```
ttpcodes, line 27: We should keep the case of the user being warned about the volume being attached. Otherwise, the user may be attaching and using the volume without Kubernetes but not get a warning since there are no workloads and encounter errors. The attached case and the workload case should be separate.

src/routes/volume/VolumeBulkActions.js
```diff
@@ -9,33 +9,32 @@ function bulkActions({ selectedRows, engineImages, bulkDeleteVolume, showBulkEng
...
           })}</div>
         </div>))
-      }
-      if (item.state === 'attached') {
-        hasVolumeAttached = true
+        hasWorkloadsRows.push(item.name)
       }
     })
     return (<div>
-      {hasVolumeAttached ? <div>Some of the selected volumes are attached to a node and may cause errors in any applications using them once they are deleted.</div> : ''}
```
ttpcodes, line 33: Same here. We can keep this message for the case of volumes being attached (even if they don't necessarily have workloads).

src/routes/volume/VolumeBulkActions.js
```diff
@@ -9,33 +9,37 @@ function bulkActions({ selectedRows, engineImages, bulkDeleteVolume, showBulkEng
...
       if (item.state === 'attached') {
-        hasVolumeAttached = true
+        hasVolumeAttachedRows.push(item.name)
       }
     })
     return (<div>
-      {hasVolumeAttached ? <div>Some of the selected volumes are attached to a node and may cause errors in any applications using them once they are deleted.</div> : ''}
-      {!hasVolumeAttached && workloadResources.length > 0 ? <div>The following applications used these volumes:</div> : '' }
+      {hasVolumeAttachedRows.length > 0 ? <div>{`Some of the selected volumes (${hasVolumeAttachedRows.map(item => item).join(', ')}) are attached to a node and may cause errors in any applications using them once they are deleted.`}</div> : ''}
+      {workloadResources.length > 0 ? <div>{ `The following volume(s) (${hasWorkloadsRows.map(item => item).join(', ')}) have workloads associated with them which may encounter errors once the volume(s) are deleted:` }</div> : '' }
```
ttpcodes, line 38: Can you change `The following volume(s)` to `Some of the selected volumes`?

PR #315, opened 2020-08-18
fix(replica) Fix the logic of replicas link passing params

smallteeths:
https://github.com/longhorn/longhorn/issues/1697

image: wangsiye/longhorn-ui:6f98076


yasker:
Signed-off-by is missing.

PR #330, opened 2020-10-13
Add drone build for ARM64 support

meldafrawi:
Signed-off-by: Mohamed Eldafrawi <mohamed.eldafrawi@rancher.com>

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,6 +1,6 @@
 ---
 kind: pipeline
-name: default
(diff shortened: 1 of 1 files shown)


boknowswiki:
@yasker , for UI repo, we don't need any changes, it has everything to support ARM64 already. We just need this drone.yaml. Thanks.

yasker (approved):

PR #351, opened 2021-01-07
fix(volume) Add detach tooltip message for RWX volume

smallteeths:
https://github.com/longhorn/longhorn/issues/2059

image:  wangsiye/longhorn-ui:03ed7dc


joshimoo (approved):
LGTM, we prohibit the user from detaching an actively in use RWX volume from the ui since it doesn't make sense to do so. If really desired the api still allows for detaching.

PR #354, opened 2021-02-09
Support raw or qcow2 file as the backing image

smallteeths:
https://github.com/longhorn/longhorn/issues/2006


shuo-wu (changes requested):
Rename commit titile to `Support raw or qcow2 file as the backing image` or `Enable backing image`

yasker (approved):
According to @smallteeths , @shuo-wu is fine with the change now.

PR #357, opened 2021-02-18
Fix readiness status when node is down

c3y1huang:
When a node is down, change its Readiness status to NotReady.

Screenshot-2021-02-18-18:52:40

https://github.com/longhorn/longhorn/issues/2186#issuecomment-765040722 - 2.


smallteeths (approved):
LGTM

innobead (approved):

PR #361, opened 2021-02-26
feat(node) Support to modify node cpu reservation

smallteeths:
https://github.com/longhorn/longhorn/issues/2207

### DIFF
--- a/src/routes/setting/setting.js
+++ b/src/routes/setting/setting.js
@@ -80,9 +80,10 @@ const form = ({
         valuePropName = 'value'
         break
     }
+    let deprecatedSettings = setting.definition && setting.definition.type === 'deprecated'
 
     return (
-      <FormItem key={setting.id} className={'settings-container'}>
+      <FormItem key={setting.id} className={'settings-container'} style={{ display: deprecatedSettings ? 'none' : 'block' }}>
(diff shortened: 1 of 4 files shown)


src/routes/setting/setting.js
```diff
@@ -80,9 +80,11 @@ const form = ({
         valuePropName = 'value'
         break
     }
+    // This part of the code will be deleted when the backend is removed and the setting is changed
+    let temporarilyHideSettings = setting.id === 'guaranteed-engine-cpu'
```
shuo-wu, line 84: You can directly check the type of this setting. The type here becomes `deprecated`:
```
{
"actions": { },
"definition": {
"category": "danger Zone",
"default": "",
"description": "This setting is replaced by 2 new settings \"Guaranteed Engine Manager CPU\" and \"Guaranteed Replica Manager CPU\" since Longhorn version v1.1.1. \nThis setting was used to control the CPU requests of all Longhorn Instance Manager pods. \n",
"displayName": "Guaranteed Engine CPU (Deprecated)",
"readOnly": true,
"required": false,
"type": "deprecated",
},
"id": "guaranteed-engine-cpu",
"links": {
"self": "…/v1/settings/guaranteed-engine-cpu",
},
"name": "guaranteed-engine-cpu",
"timestamp": "2021-03-08T06:18:23.426111167Z",
"type": "setting",
"value": "",
}
```

shuo-wu (changes requested):
Put the setting modification into a separate commit

src/routes/setting/setting.js
smallteeths: ok

shuo-wu (approved):

PR #365, opened 2021-03-10
fix(node) Remove 'Enable Scheduling' options for 'down' node

smallteeths:
https://github.com/longhorn/longhorn/issues/2308

wangsiye/longhorn-ui:6b7d26f


innobead (approved):
LGTM.

note: remove `node scheulding` from the node item menu, but only supported in `edit node` in bulk mode

PR #367, opened 2021-03-22
fix(volume) Disable the expand operation when the status of the rwx v…

smallteeths:
https://github.com/longhorn/longhorn/issues/2389

### DIFF
--- a/src/routes/volume/VolumeActions.js
+++ b/src/routes/volume/VolumeActions.js
@@ -156,6 +156,10 @@ function actions({ selected, engineImages, showAttachHost, detach, showEngineUpg
     return false
   }
 
+  const isRWXVolumeAndAttached = () => {
+    return selected.accessMode && selected.accessMode === 'rwx' && selected.state === 'attached'
+  }
+
   const upgradingEngine = () => selected.currentImage !== selected.engineImage
 
   const allActions = [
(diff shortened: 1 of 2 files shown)


innobead (approved):
LGTM

src/routes/volume/VolumeActions.js
```diff
@@ -177,7 +181,7 @@ function actions({ selected, engineImages, showAttachHost, detach, showEngineUpg
     }
   })
 
-  availableActions.push({ key: 'expandVolume', name: 'Expand Volume', disabled: !(selected.actions && (Object.keys(selected.actions).includes('attach'))) || (selected.conditions && selected.conditions.scheduled && selected.conditions.scheduled && selected.conditions.scheduled.status.toLowerCase() === 'false') })
+  availableActions.push({ key: 'expandVolume', name: 'Expand Volume', disabled: !(selected.actions && (Object.keys(selected.actions).includes('attach'))) || (selected.conditions && selected.conditions.scheduled && selected.conditions.scheduled && selected.conditions.scheduled.status.toLowerCase() === 'false') || isRWXVolumeAndAttached() })
```
joshimoo, line 184: I think you can make this more robust and simplify this further, 
expand should only be available while the selected.state === `detached`

joshimoo:
Seems that their is way to much state checking in there right now?
Can we just use `disabled: selected.state != 'detached'`

PR #368, opened 2021-03-24
fix the conditions for rwx volume to perform detach operation

smallteeths:
https://github.com/longhorn/longhorn/issues/2338


innobead (approved):

PhanLe1010 (approved):
LGTM, thanks for the fix

PR #369, opened 2021-03-24
fix(action) fix the conditions of the volume expand action

smallteeths:

### DIFF
--- a/src/routes/volume/VolumeActions.js
+++ b/src/routes/volume/VolumeActions.js
@@ -156,10 +156,6 @@ function actions({ selected, engineImages, showAttachHost, detach, showEngineUpg
     return false
   }
 
-  const isRWXVolumeAndAttached = () => {
-    return selected.accessMode && selected.accessMode === 'rwx' && selected.state === 'attached'
-  }
-
   const upgradingEngine = () => selected.currentImage !== selected.engineImage
 
   const allActions = [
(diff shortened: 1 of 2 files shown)


src/routes/volume/VolumeActions.js
```diff
@@ -181,7 +177,7 @@ function actions({ selected, engineImages, showAttachHost, detach, showEngineUpg
     }
   })
 
-  availableActions.push({ key: 'expandVolume', name: 'Expand Volume', disabled: !(selected.actions && (Object.keys(selected.actions).includes('attach'))) || (selected.conditions && selected.conditions.scheduled && selected.conditions.scheduled && selected.conditions.scheduled.status.toLowerCase() === 'false') || isRWXVolumeAndAttached() })
+  availableActions.push({ key: 'expandVolume', name: 'Expand Volume', disabled: selected.state !== 'detached' || (selected.conditions && selected.conditions.scheduled && selected.conditions.scheduled && selected.conditions.scheduled.status.toLowerCase() === 'false') })
```
innobead, line 180:
```suggestion
  availableActions.push({ key: 'expandVolume', name: 'Expand Volume', disabled: selected.state !== 'detached' || (selected.conditions && selected.conditions.scheduled && selected.conditions.scheduled.status.toLowerCase() === 'false') })
```

innobead (approved):
LGTM, some comments needed to resolve.

innobead (changes requested):
LGTM, some comments needed to resolve.

src/routes/volume/VolumeActions.js
smallteeths: Thanks for your reminding, mistakes have been corrected.

innobead (approved):

joshimoo (approved):
LGTM, as a future refactor consider creating these functions in a helper/util.
This way you only have a central place to modify whether a certain action is valid.
It also cleans up the UI code files, since isVolumeExpandable is expressive enough to communicate the intent and if the details are required a single click is all that is needed to look at them :)

```
bool isVolumeExpandable(volume) {
// do the state == detached check
// do the scheduling check
// whatever other check might be necessary
}
```

Than for the usage you can just negate that functions result
`disabled: !isVolumeExpandabale(selected)`

bulk:
`disabled: !(selectedRows.length >0 && selectedRows.all(isVolumeExpandable))`

smallteeths:
> LGTM, as a future refactor consider creating these functions in a helper/util.
> This way you only have a central place to modify whether a certain action is valid.
> It also cleans up the UI code files, since isVolumeExpandable is expressive enough to communicate the intent and if the details are required a single click is all that is needed to look at them :)
> 
> ```
> bool isVolumeExpandable(volume) {
> // do the state == detached check
> // do the scheduling check
> // whatever other check might be necessary
> }
> ```
> 
> Than for the usage you can just negate that functions result
> `disabled: !isVolumeExpandabale(selected)`
> 
> bulk:
> `disabled: !(selectedRows.length >0 && selectedRows.all(isVolumeExpandable))`

Thank you for your help. If I get a chance to refactor I will extract the public judgment methods like you said, not only modifying the `expand volume` but also the other `actions`.
longhorn-ui does needs a lot of refactoring (because it is still front-end paging and there are many unreasonable places). I'll ask yasker if he can give me some time to refactor the ui after the backend api is refactored.

PR #373, opened 2021-03-31
fix(node) Fix the status of engineimage on node

smallteeths:
https://github.com/longhorn/longhorn/issues/2377

### DIFF
--- a/src/routes/host/components/InstanceManagerComponent.js
+++ b/src/routes/host/components/InstanceManagerComponent.js
@@ -12,13 +12,12 @@ class InstanceManagerComponent extends React.Component {
     const currentNode = this.props.currentNode
 
     let data = []
-    let engineimageObj = { name: 'Engine Image', image: 'N/A', state: 'N/A', id: 'engineRowKey' }
+    let engineimageObj = { name: 'Engine Image', image: defaultEngineImage.value, state: 'not deployed', id: 'engineRowKey' }
     let instanceManagerObj = { name: 'Instance Manager', image: 'N/A', state: 'N/A', id: 'instanceRowKey' }
 
     engineimage.forEach((item) => {
-      if (defaultEngineImage.value === item.image) {
-        engineimageObj.image = item.image
-        engineimageObj.state = item.state
+      if (defaultEngineImage.value === item.image && item.nodeDeploymentMap && item.nodeDeploymentMap[currentNode.id]) {
+        engineimageObj.state = 'deployed'
       }
     })
 


src/routes/host/components/InstanceManagerComponent.js
```diff
@@ -15,11 +15,15 @@ class InstanceManagerComponent extends React.Component {
...
+    engineimage.some((item) => {
+      if (defaultEngineImage.value === item.image && item.nodeDeploymentMap && item.nodeDeploymentMap[currentNode.id]) {
         engineimageObj.image = item.image
         engineimageObj.state = item.state
+        return true
       }
+      engineimageObj.image = defaultEngineImage.value
+      engineimageObj.state = 'not deployed'
+      return false
     })
```
PhanLe1010, line 27:
```suggestion
    engineimage.forEach((item) => {
      if (defaultEngineImage.value === item.image && item.nodeDeploymentMap && item.nodeDeploymentMap[currentNode.id]) {
        engineimageObj.state = item.state
       }
     })
```

src/routes/host/components/InstanceManagerComponent.js
```diff
@@ -15,11 +15,15 @@ class InstanceManagerComponent extends React.Component {
     let engineimageObj = { name: 'Engine Image', image: 'N/A', state: 'N/A', id: 'engineRowKey' }
```
PhanLe1010, line 15:
```suggestion
    let engineimageObj = { name: 'Engine Image', image: defaultEngineImage.value, state: 'not deployed', id: 'engineRowKey' }
```

PhanLe1010 (commented):
How do you think about this suggested approach?

PhanLe1010 (approved):
LGTM
Thanks @smallteeths

innobead (approved):

src/routes/host/components/InstanceManagerComponent.js
```diff
@@ -12,13 +12,12 @@ class InstanceManagerComponent extends React.Component {
     const currentNode = this.props.currentNode
 
     let data = []
-    let engineimageObj = { name: 'Engine Image', image: 'N/A', state: 'N/A', id: 'engineRowKey' }
+    let engineimageObj = { name: 'Engine Image', image: defaultEngineImage.value, state: 'not deployed', id: 'engineRowKey' }
     let instanceManagerObj = { name: 'Instance Manager', image: 'N/A', state: 'N/A', id: 'instanceRowKey' }
 
     engineimage.forEach((item) => {
```
innobead, line 18: The question is about the existing code instead of the change. 

Why we use `.forEach` in this case? because I assume `engineimage` should only have one element of data list or we just use `.some` to check.

smallteeths: Using `some` does reduce the number of loops. I thought the engine data would not be very large, so I used `foreach`.

` engineimage.some((item) => {
      if (defaultEngineImage.value === item.image && item.nodeDeploymentMap && item.nodeDeploymentMap[currentNode.id]) {
        engineimageObj.state = 'deployed'
        return true
      }
      return false
    })
`
Thank you for the reminder I will pay attention to the use of my cycle

innobead: NP, I know your intention, just double check. Thanks.

PR #376, opened 2021-04-06
feat(backing image) Add download progress for backing image

smallteeths:
https://github.com/longhorn/longhorn/issues/2295


shuo-wu (changes requested):
Show progress bars only when the state is downloading and the UUID(key) matches in the 2 maps

innobead (approved):
LGTM

shuo-wu (approved):

PR #379, opened 2021-04-07
fix(volume) fix the conditions for volume attach and detach operations

smallteeths:
https://github.com/longhorn/longhorn/issues/2411

### DIFF
--- a/src/routes/volume/helper/index.js
+++ b/src/routes/volume/helper/index.js
@@ -269,3 +269,34 @@ export function disabledSnapshotAction(volume, modelState) {
 export function extractImageVersion(image) {
   return image.substr(image.lastIndexOf(':') + 1, image.length)
 }
+
+const isRestoring = (volume) => {
+  if (volume.restoreStatus && volume.restoreStatus.length > 0) {
+    let flag = volume.restoreStatus.every((item) => {
+      return !item.isRestoring
+    })
+    return !flag
+  } else {
+    return false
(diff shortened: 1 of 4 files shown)


innobead (approved):
LGTM

src/routes/volume/helper/index.js
```diff
@@ -269,3 +269,34 @@ export function disabledSnapshotAction(volume, modelState) {
...
+  if (volume.standby || isRestoring(volume)) {
+    return false
+  }
+  if (volume.accessMode === 'rwo') {
+    return volume.state === 'attached'
+  }
+  if (volume.migratable) {
+    return volume.state === 'attached' && volume.controllers && volume.controllers.length <= 1
+  }
+  return volume.state === 'attached' && volume.disableFrontend
```
joshimoo, line 294: Might be worth flipping this, so you have conditions for RWX instead of RWO, since the RWO condition is simple, it makes it easier to understand the difference at a glance.
```
  if (volume.accessMode === 'rwx') {
    if (volume.migratable) {
      return volume.state === 'attached' && volume.controllers && volume.controllers.length <= 1
    }
    return volume.state === 'attached' && volume.disableFrontend
  }

  return volume.state === 'attached'
```

joshimoo (commented):
LGTM, left some suggestions.
Great job on the helper refactor, makes it much easier to read and update these conditions.

PR #380, opened 2021-04-07
fix(code) Optimize detach action code

smallteeths:
https://github.com/longhorn/longhorn/issues/2411

### DIFF
--- a/src/routes/volume/VolumeBulkActions.js
+++ b/src/routes/volume/VolumeBulkActions.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { Button, Modal, Menu, Dropdown, Icon } from 'antd'
-import { detachable, attachable } from './helper'
+import { detachable, attachable, isRestoring } from './helper'
 import style from './VolumeBulkActions.less'
 
 const confirm = Modal.confirm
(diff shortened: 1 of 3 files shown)


src/routes/volume/VolumeBulkActions.js
```diff
@@ -114,15 +114,8 @@ function bulkActions({ selectedRows, engineImages, bulkDeleteVolume, showBulkEng
...
-      })
-      return !flag
-    } else {
-      return false
-    }
+  const hasVolumeRestoring = () => selectedRows.some((selected) => {
+    return isRestoring(selected)
   })
   const isAutomaticallyUpgradeEngine = () => {
     return selectedRows.some((item) => {
```
joshimoo, line 121: In the future we can do some more refactoring, by moving the single case into the helper.
We might want to just have a single function exported `canUpgradeEngineImage`
This refactor is not necessary for this PR, just something to think about.
example:
```
export func canUpgradeEngine(volume, image, defaultImage, autoUpgradeLimit) {
// we pass in the required parameters for determination, this way this can easily be unit tested and helper does not
// require knowledge of the surrounding context.
// image would be the desired image
// defaultImage would be string name of the image
// autoUpgradeLimit would be int number
}
```

joshimoo (approved):
LGTM, haven't checked all the conditions, but since it's mostly a refactor, I assume you verified the conditions?

PR #381, opened 2021-04-12
fix(node) Steps for synchronous bulk deletion and individual node del…

smallteeths:
https://github.com/longhorn/longhorn/issues/2462


innobead (approved):
LGTM

PR #382, opened 2021-04-14
fix(node) Add warning message when deleting node

smallteeths:
https://github.com/longhorn/longhorn/issues/2462


innobead (approved):
LGTM

PR #383, opened 2021-04-15
fix(backingImage) URL typo error

smallteeths:
https://github.com/longhorn/longhorn/issues/2494


innobead (approved):
LGTM

PR #384, opened 2021-04-15
fix(backing) Show the backing image size

smallteeths:
https://github.com/longhorn/longhorn/issues/2497


innobead (approved):
LGTM

PR #394, opened 2021-05-24
fix(detach) Fix the detach operation in volume details page

smallteeths:
https://github.com/longhorn/longhorn/issues/2587

### DIFF
--- a/src/routes/volume/detail/index.js
+++ b/src/routes/volume/detail/index.js
@@ -22,12 +22,13 @@ import ChangeVolumeModal from '../ChangeVolumeModal'
 import ExpansionVolumeSizeModal from '../ExpansionVolumeSizeModal'
 import Salvage from '../Salvage'
 import { ReplicaList, ExpansionErrorDetail } from '../../../components'
+import ConfirmModalWithWorkload from '../ConfirmModalWithWorkload'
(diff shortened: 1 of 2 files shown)


khushboo-rancher:
Checked with the image `wangsiye/longhorn-ui:5eafaa9`, working as expected, the detach button from volume details page is working.

src/routes/volume/detail/index.js
```diff
@@ -462,6 +468,32 @@ function VolumeDetail({ snapshotModal, dispatch, backup, engineimage, eventlog,
...
+          payload: {
+            url: selectedVolume.actions.detach,
+          },
+        })
+      }
+      dispatch({
+        type: 'volume/hideConfirmDetachWithWorkload',
+      })
+      dispatch({
+        type: 'volume/stopPolling',
```
innobead, line 487: Do we have this callback? I did not see it in models/volume.js

innobead (changes requested):

src/routes/volume/detail/index.js
smallteeths: Thank you for your careful inspection, I did make a mistake in this position. After we detach the volume in the details page, we need to stop requesting snapshotdata. Otherwise, the back-end will response an error. So we need to call snapshotModal/stopPolling.

It’s okay without this callback, but it’s more rigorous if there is one.

innobead (approved):

innobead:
@smallteeths thanks! backport is needed for 1.1.2.

PR #406, opened 2021-06-10
fix(eventlog) Support toggle to enable the event log ws

smallteeths:
https://github.com/longhorn/longhorn/issues/2592


innobead:
For now, we will not go this way, let's just close the PR first, but see if we will need in the future.

PR #407, opened 2021-06-15
feat(volume) Implement Replica Auto Balance Settings

smallteeths:
https://github.com/longhorn/longhorn/issues/2676

### DIFF
--- a/src/models/volume.js
+++ b/src/models/volume.js
@@ -50,6 +50,7 @@ export default {
     updateAccessModeModalVisible: false,
     updateBulkAccessModeModalVisible: false,
     confirmModalWithWorkloadVisible: false,
+    updateReplicaAutoBalanceModalVisible: false,
     changeVolumeActivate: '',
     defaultPvOrPvcName: '',
     defaultNamespace: '',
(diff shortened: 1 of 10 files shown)


c3y1huang:
Clicking on the action dropdown `Update Replica Auto Balance` inside a specific volume seems to do no-op, but worked on the volume list page though.
screenshot-1

c3y1huang:
And noticed UpdateReplicaCount is actually `Update Replica**s** Count`, I think it will look nicer if we move the replica auto-balance just below the update replicas count and also make the wordings consistent as `Update Replica**s** Auto Balance`?

c3y1huang:
Can we also have the `Update Replica Auto Balance` for bulk action with the other `Update` actions?
Screenshot-2021-06-16-10:36:13

smallteeths:
> Can we also have the `Update Replica Auto Balance` for bulk action with the other `Update` actions?
> Screenshot-2021-06-16-10:36:13

Yes, batch operations can be performed.

c3y1huang (approved):
LGTM

innobead (approved):
LGTM

PR #414, opened 2021-06-30
[WIP] fix(axios) Upgrade axios to v0.21.1

smallteeths:


innobead (approved):

bin/latest_image
```diff
@@ -0,0 +1 @@
+longhornio/longhorn-ui:abbabef
```
innobead, line 1: I don't think we need this?

innobead (changes requested):

innobead (approved):

PR #415, opened 2021-07-01
fix(backup) Fix the backup details page data becomes empty after request

smallteeths:
https://github.com/longhorn/longhorn/issues/2754

### DIFF
--- a/src/models/backup.js
+++ b/src/models/backup.js
@@ -25,6 +25,7 @@ export default {
     diskTags: [],
     bulkRestoreData: [],
     backupVolumesForBulkCreate: [],
+    search: {},
     isBulkRestore: false,
     restoreBackupModalVisible: false,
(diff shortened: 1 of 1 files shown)


src/models/backup.js
```diff
@@ -326,6 +338,12 @@ export default {
     changeTagsLoading(state, action) {
       return { ...state, ...action.payload }
     },
+    ensureRouteLocation(state, action) {
```
innobead, line 341: Do we need this?

innobead (changes requested):

src/models/backup.js
smallteeths: This method should indeed be deleted. This is the code left over when I try to modify it in another way. Thanks.

innobead (approved):

PR #417, opened 2021-07-02
Frequently update installed packages of image

innobead:
ref: https://ci.longhorn.io/job/private/job/image-security-scan/355/

Signed-off-by: David Ko <dko@suse.com>


innobead:
cc @meldafrawi

PR #419, opened 2021-07-09
Change color scheme to match documentation and branding

peterolivosuse:
I was flipping around between Harvester, Rancher and Longhorn and noticed that the Longhorn UI was different than the branding and documentation. Did a quick color replacement and came up with this. Let me know if I need to do anything differently to shape things up.

Thanks!


smallteeths (approved):
LGTM

innobead:
Hold to merge

PR #422, opened 2021-08-06
feat(volume) UI add volume encryption support

smallteeths:
image: wangsiye/longhorn-ui:e68aeb2

https://github.com/longhorn/longhorn/issues/2841

### DIFF
--- a/src/routes/backup/RestoreBackup.js
+++ b/src/routes/backup/RestoreBackup.js
@@ -116,6 +116,12 @@ const modal = ({
             { backingImages.map(backingImage => <Option key={backingImage.name} value={backingImage.name}>{backingImage.name}</Option>) }
           </Select>)}
         </FormItem>
+        <FormItem label="Encrypted" {...formItemLayout}>
+          {getFieldDecorator('encrypted', {
+            valuePropName: 'encrypted',
+            initialValue: false,
+          })(<Checkbox></Checkbox>)}
+        </FormItem>
         <Spin spinning={tagsLoading}>
           <FormItem label="Node Tag" hasFeedback {...formItemLayout}>
             {getFieldDecorator('nodeSelector', {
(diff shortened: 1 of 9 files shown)


src/routes/backup/RestoreBackup.js
```diff
@@ -116,6 +116,12 @@ const modal = ({
             { backingImages.map(backingImage => <Option key={backingImage.name} value={backingImage.name}>{backingImage.name}</Option>) }
           </Select>)}
         </FormItem>
+        <FormItem label="Encrypted" {...formItemLayout}>
+          {getFieldDecorator('encrypted', {
+            valuePropName: 'encrypted',
+            initialValue: false,
```
innobead, line 122: Just a check. It seems users need to specify the secret info again when restoring a backup, right? (maybe there are some place to improve this part in the future)

innobead (approved):
LGTM. Just one question about restoring a backup.

I leave to @joshimoo check again before merging.

src/routes/backup/RestoreBackup.js
smallteeths: Yes, if we need to get the corresponding volume.encrypted this needs to be considered a bit more. If the volume has been deleted we won't get the encrypted property. I can consider setting the initial value for it if needed.

joshimoo (approved):
LGTM at the moment we don't store the encrypted flag as part of a backup.
So a restore requires manually specifying that this is an encrypted volume.
We also need to manually specify the secret for the PV ref when manually creating a PV for a restored volume. 

In the future we can include that encrypted field or store the kubernetes PV as part of a volume backup. To help automate some of these things.

Alternatively allow specifying a storage class for a restored volume, that way we would get all necessary information for the volume context from the storage class. 

One important thing that we cannot store is the actual secret content, since then we would be exposing the encryption keys for a volume.

PR #423, opened 2021-08-15
feat(Recurring Job) UI support Recurring Job

smallteeths:
https://github.com/longhorn/longhorn/issues/467


innobead:
cc @c3y1huang

innobead:
@c3y1huang gave some updates to me. This PR will get ready after some minor fixes later on and expect in the RC1.

c3y1huang:
Tested build, LGTM.

src/services/volume.js
```diff
@@ -119,3 +111,52 @@ export async function cancelExpansion(params) {
...
+
+export async function removeVolumeRecurringJob(params) {
+  return request({
+    url: params.url,
+    method: 'post',
+    data: {
+      name: params.name,
+      isGroup: params.isGroup,
+    },
+  })
```
innobead, line 144: This is the same as the `recurringJobAdd`? I thought it should be delete action.

innobead (changes requested):

src/services/volume.js
smallteeths: They call the url is actually not the same, delete call is selectedVolume.actions.recurringJobDelete, the parameters are the same. I can combine them but the names of the functions will be very strange.

I wanted to separate `delete` and `add` so I used these two functions.

innobead (approved):
LGTM

PR #432, opened 2021-08-27
fix(nginx) Adjusting nginx proxy send time

smallteeths:
https://github.com/longhorn/longhorn/issues/2937

image: wangsiye/longhorn-ui:d6e1099


innobead:
Let's close this first because the original question resolved and increasing the timeout is not what we pursue at the moment.

PR #435, opened 2021-09-13
fix (recurringJob)

smallteeths:
https://github.com/longhorn/longhorn/issues/2996

1. Disable the add button when the volume has no actions associated with the recurringjob
2. Does not determine if the volume has been bound to a recurring job only displays the icon

### DIFF
--- a/src/models/volume.js
+++ b/src/models/volume.js
@@ -124,7 +124,9 @@ export default {
     *queryVolumeRecurringJobList({
       payload,
     }, { call, put }) {
-      if (payload.actions) {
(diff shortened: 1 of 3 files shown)


innobead (approved):
LGTM

@smallteeths Please backport to 1.2.x.

Note: This change will allow users to click the schedule icon anyway to launch the recurring job modal window to do the job management. This is different than the previous behavior before the new recurring jobs were introduced in 1.2.0. If need to determine the recurring jobs of each volume in the volume page, need backend extra work to provide the info instead.

PR #442, opened 2021-09-22
feat(footer) Add Stable Longhorn Versions information display

smallteeths:
https://github.com/longhorn/longhorn/issues/3032

image: wangsiye/longhorn-ui:f696a77

### DIFF
--- a/src/components/Layout/Footer.js
+++ b/src/components/Layout/Footer.js
@@ -8,15 +8,18 @@ import { getStatusIcon, getBackupStatusIcon } from '../../utils/websocket'
 import upgradeIcon from '../../assets/images/upgrade.svg'
 import semver from 'semver'
 import BundlesModel from './BundlesModel'
+import StableLonghornVersions from './StableLonghornVersions'
 
 function Footer({ app, host, volume, setting, engineimage, eventlog, backingImage, recurringJob, backup, dispatch }) {
-  const { bundlesropsVisible, bundlesropsKey, okText, modalButtonDisabled, progressPercentage } = app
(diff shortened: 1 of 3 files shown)


src/components/Layout/Footer.js
```diff
@@ -80,6 +113,7 @@ function Footer({ app, host, volume, setting, engineimage, eventlog, backingImag
           <a target="blank" onClick={showBundlesModel}>Generate Support Bundle</a>
           <a target="blank" href={issueHref}>File an Issue</a>
           <a target="blank" href="https://slack.cncf.io/">Slack</a>
+          <a target="blank" onClick={showStableLonghornVersions}>{`Stable Longhorn Versions (${stableLonghornVersions})`}</a>
```
PhanLe1010, line 116: Can we change the text `Stable Longhorn Versions` to `Newer Stable Versions` ?

PhanLe1010 (changes requested):
We only want to show the stable versions that are newer than the current version. If there is no newer stable version, we don't show the `stable versions` list on the bottom of the page.

PhanLe1010 (approved):
LGTM. Thanks @smallteeths !

innobead (approved):
LGTM, but it seems we can't do the filtering of the latest stable versions in the setting page but do in the footer.

But, It's good for now.

innobead:
@smallteeths Please help backport to 1.2.x.

PhanLe1010:
@innobead Can we also backport to v1.1.3 ?

innobead:
> @innobead Can we also backport to v1.1.3 ?

We should. :)

PR #446, opened 2021-09-29
Do not copy Host header to API requests forwarded to Longhorn Manager

dagan:
Longhorn UI serves API requests (/v1/**) via reverse proxy to the Longhorn Manager:

```
location /v1  {
    proxy_pass ${LONGHORN_MANAGER_IP};
    proxy_request_buffering off;
    client_max_body_size 10240M;
}
```

For forwarded requests, the Host header is copied from the original request:

```
proxy_set_header   Host             $http_host;
```

Although the default NGINX behavior is to set Host to the upstream host, overriding this header with the proxy host as longhorn-ui does is not uncommon (it can be useful for constructing absolute URLs that will route via the proxy, for example). However, when Longhorn is deployed in a cluster with a service mesh using Envoy as a sidecar, setting this header breaks routing to Longhorn Manager.

For example, in an Istio mesh, the longhorn-ui sidecar will have a listener configured for routing traffic from longhorn-ui to longhorn-backend (manager) that is based on the longhorn-backend service (i.e., {LONGHORN_BACKEND_CLUSTER_IP}:9500). The listener will support the various hostnames for the service (longhorn-backend, longhorn-backend.longhorn-system, etc.). It will not, however, support the Host used to access the longhorn-ui service. In this scenario, API requests will fail as follows:

 1. API requests arrive at longhorn-ui and match the /v1 location
 2. The Host header value from the inbound request (the Host used to reach longhorn-ui) is copied to the proxied request
 3. DNS lookup for the longhorn-backend service occurs, resolving the service's CLUSTER_IP
 4. The proxied HTTP request is sent to the longhorn-backend CLUSTER_IP with a Host header matching the request received by longhorn-ui (e.g., dashboard.rancher.example.com for Rancher clusters).
 5. The request is intercepted by Envoy and matched to the longhorn-backend service listen based on the destination IP and port.
 6. Envoy is not able to identify a route for the request because the Host header value does not match one of the expected hosts.
 7. Envoy rejects the request with a 404 status code.

This PR reverts the Host header to NGINX's default behavior of setting the Host header to the destination hostname (e.g. longhorn-backend), allowing Envoy to find the appropriate route and forward the request.

Alternatively, X-Forwarded-Host can be used retain information about the host used for original request.

Signed-off-by: Dagan Henderson <dhenderson@goraft.tech>

### DIFF
--- a/nginx.conf.template
+++ b/nginx.conf.template
@@ -15,7 +15,6 @@ http {
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection "upgrade";
         proxy_redirect     off;
-        proxy_set_header   Host             $http_host;
         proxy_set_header   X-Real-IP        $remote_addr;
         proxy_set_header   X-Forwarded-For  $proxy_add_x_forwarded_for;
         proxy_next_upstream error timeout invalid_header http_500 http_502 http_503 http_504;


smallteeths:
The action address returned by the backend is based on the host as shown below. If we need to be compatible we also need to configure `proxy_set_header X-Forwarded-Host $host` in nginx. Then the backend needs to be modified as well. The action url returned by the backend needs to be stitched together according to `X-Forwarded-Host`. @dagan @PhanLe1010 cc @innobead 

截屏2021-12-02 上午11 24 31

innobead:
@smallteeths 

Could you try `X-Forwarded-Host` instead of overriding `Host` to see if it will work w/o any backend change? From the code, that should just work as expected, because the forward header will take precedence over the original host header. 

[/vendor/github.com/rancher/go-rancher/api/url.go#L158-L191](https://github.com/longhorn/longhorn-manager/blob/d5498b1dcf88f56ade7f9ce17a4c75cb3aae47a3/vendor/github.com/rancher/go-rancher/api/url.go#L158-L191)
```go
func getUrlFromStandardHeaders(r *http.Request) string {
	xForwardedProto := getOverrideHeader(r, FORWARDED_PROTO_HEADER, "")
	if xForwardedProto == "" {
		return ""
	}

	host := getOverrideHeader(r, FORWARDED_HOST_HEADER, "")
	if host == "" {
		host = r.Host
	}

	if host == "" {
		return ""
	}

	port := getOverrideHeader(r, FORWARDED_PORT_HEADER, "")
	if port == "443" || port == "80" {
		port = "" // Don't include default ports in url
	}

	if port != "" && strings.Contains(host, ":") {
		// Have to strip the port that is in the host. Handle IPv6, which has this format: [::1]:8080
		if (strings.HasPrefix(host, "[") && strings.Contains(host, "]:")) || !strings.HasPrefix(host, "[") {
			host = host[0:strings.LastIndex(host, ":")]
		}
	}

	if port != "" {
		port = ":" + port
	}

	return fmt.Sprintf("%s://%s%s%s", xForwardedProto, host, port, r.URL.Path)
}
```

innobead:
Let's follow up at https://github.com/longhorn/longhorn-ui/pull/596 instead. 

Thanks @dagan for the contribution :+1:

PR #452, opened 2021-10-14
Add true random number generator. Replace Math.Random().

100mik:
Resolves `opt.semgrep.node_insecure_random_generator` observed though [Sonatype Lift](https://lift.sonatype.com/result/longhorn/longhorn-ui/01FHW2P5PASS9XVK4N1VXG2SGR?t=CustomTool%20%22Semgrep%22%7Copt.semgrep.node_insecure_random_generator)


innobead:
Thanks for the contribution! 

For this PR, I am concerned about the browser backward compatibility and math.random usage only for the identity key of UI components, so it will be kept to us on purpose.

Closing, but feel free to contribute again if you think this is really helpful.

PR #453, opened 2021-10-28
feat(cron) Fix the initial value of cron expression when editing Recu…

smallteeths:
https://github.com/longhorn/longhorn/issues/3180

image: wangsiye/longhorn-ui:5c4d086


innobead (approved):
LGTM

PR #458, opened 2021-11-05
fix(axios) Fix the request address path is incorrectly spliced when t…

smallteeths:
image: wangsiye/longhorn-ui:ee75d7a

https://github.com/longhorn/longhorn/issues/3191


innobead (approved):
LGTM

PR #471, opened 2021-11-16
fix(RecurringJob) Fix the incorrect cron expression display in volume…

smallteeths:
https://github.com/longhorn/longhorn/issues/3180

image: wangsiye/longhorn-ui:a91ab99


innobead (approved):
LGTM

PR #473, opened 2021-11-26
feat(volume) Add the engine image field to the volume list

smallteeths:

image: wangsiye/longhorn-ui:dd8637b

截屏2021-12-01 下午5 05 20

截屏2021-12-01 下午5 05 51

截屏2021-12-01 下午5 06 00

### DIFF
--- a/src/publicPath.js
+++ b/src/publicPath.js
@@ -17,6 +17,7 @@ let columnArr = [
 	'recurringJobs',
 	'lastBackupAt',
 	'actualSize',
+	'engineImage',
 ]
 let pageSizeCollectionObject = {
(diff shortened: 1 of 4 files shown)


innobead:
@smallteeths Could we have a screenshot? Also fix this in the master branch first.

smallteeths:
> @smallteeths Could we have a screenshot? Also fix this in the master branch first.

ok

joshimoo:
cc @innobead can you clarify, I don't think this is what was requested?

innobead:
> cc @innobead can you clarify, I don't think this is what was requested?

This is not like you guys discussed, but instead instance manager image.

PR #474, opened 2021-11-30
fix(setting) Fix revision counter setting typo error

smallteeths:
https://github.com/longhorn/longhorn/issues/3308


innobead (approved):
LGTM

Please also fix in the master branch.

PR #479, opened 2021-12-06
feat(volume) Add the engine image field to the volume list

smallteeths:


innobead:
cc @kaxing please review this from UI POV.

kaxing:
A curious question, does "Engine Image" column going to be shown by default?

smallteeths:
> A curious question, does "Engine Image" column going to be shown by default?

Yes it will be displayed by default

PR #480, opened 2021-12-09
feat(instance manager) Support for display instance manager image inf…

smallteeths:

### DIFF
--- a/src/models/backingImage.js
+++ b/src/models/backingImage.js
@@ -1,9 +1,9 @@
 import { create, deleteBackingImage, query, deleteDisksOnBackingImage, uploadChunk } from '../services/backingImage'
-import { parse } from 'qs'
 import { message, notification } from 'antd'
 import { delay } from 'dva/saga'
 import { wsChanges, updateState } from '../utils/websocket'
 import queryString from 'query-string'
+import { enableQueryData } from '../utils/dataDependency'
 
 export default {
   ws: null,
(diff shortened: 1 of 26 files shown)


src/models/backingImage.js
```diff
@@ -26,10 +27,12 @@ export default {
...
     setup({ dispatch, history }) {
       history.listen(location => {
-        dispatch({
-          type: 'query',
-          payload: location.pathname === '/backingImage' ? queryString.parse(location.search) : '',
-        })
+        if (enableQueryData(location.pathname, 'backingImage')) {
+          dispatch({
+            type: 'query',
+            payload: location.pathname.startsWith('/backingImage') ? queryString.parse(location.search) : {},
```
innobead, line 33: `enableQueryData` already does the pathname check. Do we still need this ternary expression? I think we just need `payload: queryString.parse(location.search)`

The same for others.

src/models/dashboard.js
```diff
@@ -9,18 +10,20 @@ export default {
...
+        }
       })
     },
   },
   effects: {
     *query({
       payload,
     }, { call, put }) {
-      const data = yield call(query, parse(payload))
+      const data = yield call(query, queryString.parse(payload))
```
innobead, line 26: Not sure if we still need parse the payload, because it has been parsed before dispatch as above. 

If need to change, please change other places as well.

innobead (changes requested):
In general, LGTM. Few items needed to clarify.

src/models/dashboard.js
smallteeths: In fact, this parameter will not be wrong even if I don't pass it. Now all the logic of the data filter is done by the front end. Even if I pass the parameter to the backend through the interface, the backend will return the full amount of data. You are right, I should optimize this piece of code for future support of back-end filtering

src/models/backingImage.js
smallteeths: `enableQueryData` is only used to determine which http request the current page needs to load. It doesn't determine if the current request needs to send parameters to filter data. For example, In the volume page we needs to load node data and volume data. Because it’s in volume page so only the volume data needs to be filtered, node data don't need to be filtered. So I need this judgment to get the full node data.

innobead: You're right 👍

innobead (approved):

PR #488, opened 2022-01-07
fix(volume) Fix volume list state when volume.robustness is unknown

smallteeths:
https://github.com/longhorn/longhorn/issues/3291

image: wangsiye/longhorn-ui:78c0fad

The pr mainly fixes the problem that the volume list and volume deatil page states are different when the volume state is attached and robustness is unkonw.


innobead (approved):
LGTM

PR #489, opened 2022-01-07
Support display instance manager image list

smallteeths:
image: wangsiye/longhorn-ui:398a886

The pr implements two feature.
1. Load http requests on demand. The original longhorn-ui loads apis with all the data once every time the page is switched. This doesn't make much sense. It should load http requests on demand to reduce the pressure on the server.
2. Support the display of instance manager image list.

### DIFF
--- a/src/models/dashboard.js
+++ b/src/models/dashboard.js
@@ -1,5 +1,6 @@
 import { query } from '../services/dashboard'
-import { parse } from 'qs'
+import queryString from 'query-string'
+import { enableQueryData } from '../utils/dataDependency'
 
 export default {
   namespace: 'dashboard',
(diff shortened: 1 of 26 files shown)


src/models/dashboard.js
```diff
@@ -9,18 +10,20 @@ export default {
...
+        }
       })
     },
   },
   effects: {
     *query({
       payload,
     }, { call, put }) {
-      const data = yield call(query, parse(payload))
+      const data = yield call(query, queryString.parse(payload))
```
innobead, line 26: Please sync with the PR from the master.

innobead (changes requested):

innobead (approved):

PR #490, opened 2022-01-12
fix(volume) Fix volume list state when volume.robustness is unknown

smallteeths:
https://github.com/longhorn/longhorn/issues/3291


innobead (approved):

innobead:
@smallteeths Also fix in the master branch

PR #495, opened 2022-02-24
Moving to NGINX to tempfs

mattmattox:
This moves the NGINX config files to a tempfs because in a hardened environment root is a read-only system so we need to move it to an emptyDir that is mounted with fsGroup being set.


smallteeths (approved):
LGTM

Jellyfrog:
This is great, but maybe consider using the "unprivileged" image instead, https://hub.docker.com/r/nginxinc/nginx-unprivileged 
It is an offical image, just under a different account for some reason; https://github.com/nginxinc/docker-nginx-unprivileged/issues/19

PR #500, opened 2022-04-02
[Master] feat(snapshot) Allows users to delete a snapshot that has been marked…

smallteeths:
https://github.com/longhorn/longhorn/issues/3748

### DIFF
--- a/src/routes/volume/detail/Snapshots.js
+++ b/src/routes/volume/detail/Snapshots.js
@@ -44,21 +44,34 @@ class Snapshots extends React.Component {
         })
         return
       }
-      let actions
+
+      let additionalActions = []
+      let actionType = action.type
+      let url = this.props.volume.actions[action.type]
+      let params = {
+        name: action.payload && action.payload.snapshot ? action.payload.snapshot.name : undefined,
+      }
+
       if (action.type === 'snapshotDelete') {
-        actions = [{
+        additionalActions = [{
           url: this.props.volume.actions.snapshotPurge,
         }]
+        // Only snapshotPurge operations are performed when the snapshot status is marked as removed.
+        if (action.payload?.snapshot?.removed) {
+          additionalActions = []
+          actionType = 'snapshotPurge'
+          url = this.props.volume.actions?.snapshotPurge
+          params = {}
+        }
       }
+
       this.props.dispatch({
         type: 'snapshotModal/snapshotAction',
         payload: {
-          type: action.type,
-          actions,
-          url: this.props.volume.actions[action.type],
-          params: {
-            name: action.payload && action.payload.snapshot ? action.payload.snapshot.name : undefined,
-          },
+          type: actionType,
+          actions: additionalActions,
+          url,
+          params,
           querySnapShotUrl: this.props.volume.actions.snapshotList,
         },
       })
--- a/src/components/Snapshot/Snapshot.js
+++ b/src/components/Snapshot/Snapshot.js
@@ -86,19 +86,20 @@ function SnapshotIcon(props, snapshotProps) {
     doAction(key)
   }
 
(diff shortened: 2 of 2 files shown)


src/routes/volume/detail/Snapshots.js
```diff
@@ -45,23 +45,34 @@ class Snapshots extends React.Component {
...
-      }
-      this.props.dispatch({
-        type: 'snapshotModal/snapshotAction',
-        payload: {
-          type: action.type,
-          actions,
-          url: this.props.volume.actions[action.type],
-          params: {
-            name: action.payload && action.payload.snapshot ? action.payload.snapshot.name : undefined,
+      if (!action.payload?.snapshot?.removed) {
```
innobead, line 48: Do we need this condition? Because whatever the value of `removed` is, we always need to convert to use `snapshotPurge` with `snapshotDelete`.

Also, this chang will cause other actions instead of `snapshotBackup`, `toggleShowRemoved` and `backup` converted to `snapshotPurge` when `snapshot.removed` is true.

I feel we should keep the case as is. WDYT?

innobead (changes requested):

src/routes/volume/detail/Snapshots.js
smallteeths: Sorry that this code is really misleading. The normal delete snapshot actually executes the request snapshotDelete+snapshotPurge twice. The corresponding called function will determine whether the actions array still has the requested link, and if there is, the request will continue.

截屏2022-04-22 下午4 10 30

To delete a snapshot that has been marked as removed, simply execute snapshotPurge. So this judgement is still needed.

I do need to optimize this code to make my intentions more explicit. I will adjust this pr to WIP. I'll let you know when the optimization is complete. Thank you for your patience in checking. @innobead

innobead: ah, thanks for the explanation. Waiting for the update then.

src/routes/volume/detail/Snapshots.js
```diff
@@ -44,24 +44,34 @@ class Snapshots extends React.Component {
...
+            actions: additionalActions,
+            url,
+            params,
+            querySnapShotUrl: this.props.volume.actions.snapshotList,
           },
-          querySnapShotUrl: this.props.volume.actions.snapshotList,
-        },
-      })
+        })
+      }
```
innobead, line 74:
```js
	  let actionType = action.type
	  let url = this.props.volume.actions[action.type]
	  let params = {
	    name: action.payload && action.payload.snapshot ? action.payload.snapshot.name : undefined,
	  }

	  if (action.type === 'snapshotDelete') {
		    let additionalActions = [{
		      url: this.props.volume.actions.snapshotPurge,
		    }]
	  
		    // Only snapshotPurge operations are performed when the snapshot status is marked as removed.
		    if (action.payload?.snapshot?.removed) {
		      additionalActions = []
		      actionType = 'snapshotPurge'
		      url = this.props.volume.actions?.snapshotPurge
		      params = {}
		    }
		}
	  
	  this.props.dispatch({
	    type: 'snapshotModal/snapshotAction',
	    payload: {
	      type: actionType,
	      actions: additionalActions,
	      url,
	      params,
	      querySnapShotUrl: this.props.volume.actions.snapshotList,
	    },
	  })
```

smallteeths: I just checked again that your comments is correct. I had thought that all other action.types were returned. But I double-checked action.revet , which does execute the default `snapshotModal/snapshotAction`.  Thank you for your patience in checking ！

innobead (approved):

innobead:
This should be backported to 1.2.x. https://github.com/longhorn/longhorn/issues/3768 

cc @shuo-wu

PR #501, opened 2022-04-08
feat(nginx) Make Longhorn UI Listen Binding configurable

smallteeths:
https://github.com/longhorn/longhorn/issues/3829

image: wangsiye/longhorn-ui:98a4dd2

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,10 +23,10 @@ WORKDIR /web
 COPY --from=builder /web/dist /web/dist
 COPY --from=builder /web/nginx.conf.template /etc/nginx/nginx.conf.template
 
-EXPOSE 8000
 ENV LONGHORN_MANAGER_IP http://localhost:9500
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -23,10 +23,10 @@ WORKDIR /web
 COPY --from=builder /web/dist /web/dist
 COPY --from=builder /web/nginx.conf.template /etc/nginx/nginx.conf.template
 
-EXPOSE 8000
 ENV LONGHORN_MANAGER_IP http://localhost:9500
+ENV LONGHORN_UI_LISTEN_PORT 8000
```
innobead, line 27: Suggest using LONGHORN_UI_PORT

innobead (changes requested):
Please also update longhorn/longhorn and longhorn-manager chart manifests as well.

innobead:
@smallteeths Let's make this change to the master only and no need to backport to 1.2.x.

PR #502, opened 2022-04-08
fix(PV/PVC) Disable PV/PVC creation while volume is being created

smallteeths:
image: wangsiye/longhorn-ui:54b0421 (v1.2.x)

https://github.com/longhorn/longhorn/issues/3807


innobead (approved):
LGTM

innobead:
ref: http://github.com/longhorn/longhorn-manager/blob/01c0e8c3f278a26e104c8973ee3321bd7c3c3d73/api/model.go#L1128-L1166

innobead:
@smallteeths also need to fix in the master branch.

PR #510, opened 2022-04-14
Add s390x support

thomasferrandiz:
This PR adds support to build on s390x architecture.

PR #511, opened 2022-04-18
change base image to SUSE BCI

meldafrawi:
Signed-off-by: Mohamed Eldafrawi <mohamed.eldafrawi@suse.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,23 +10,28 @@ ENV VERSION ${VERSION}
 RUN envsubst '${VERSION}' < /web/src/utils/config.js > /web/src/utils/config.js.subst && mv /web/src/utils/config.js.subst /web/src/utils/config.js
 RUN npm run build
 
-FROM nginx:1.20.1-alpine
-RUN apk update && \
-    apk upgrade && \
-    apk add curl && \
-    apk add bash && \
-    apk add gettext && \
-    apk add shadow && \
-    rm -rf /var/cache/apk/*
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -10,23 +10,28 @@ ENV VERSION ${VERSION}
...
 WORKDIR /web
+
 COPY --from=builder /web/dist /web/dist
 COPY --from=builder /web/nginx.conf.template /etc/nginx/nginx.conf.template
 
+EXPOSE 8000
 ENV LONGHORN_MANAGER_IP http://localhost:9500
 ENV LONGHORN_UI_PORT 8000
 
-RUN useradd -r -u 1000 longhorn
```
innobead, line 29: Why we removed this? This is for security hardened.

innobead (changes requested):

Dockerfile
meldafrawi: Nginx will start using `nginx`  user

innobead: It looks you are using nginx rpm package to install, but unfortunately nginx doesn't support arm64. Also, this will be a problem when we support other architectures. Probably we should just use the official niginx image instead.

meldafrawi: Sure, I'll close the PR, and we will continue using Nginx official image.

innobead: https://github.com/longhorn/longhorn-ui/pull/510#issuecomment-1103325818

Let's see if we can use Nginx from the built-in repo in the BCI image directly.

PR #512, opened 2022-04-21
change base image to SUSE BCI

meldafrawi:
Signed-off-by: Mohamed Eldafrawi <mohamed.eldafrawi@suse.com>

Ref: https://github.com/longhorn/longhorn/issues/2939


innobead (approved):

smallteeths:
The original reason for using the `nginx:1.20.1-alpine` image was to reduce the size of the longhorn-ui front-end image.

 `nginx:1.20.1-alpine`: 22.7mb
`registry.suse.com/bci/bci-base`: 116mb

Replacing with `registry.suse.com/bci/bci-base` will not increase too much. LGTM @innobead

PR #516, opened 2022-05-06
feat(orphaned) Add orphaned data page

smallteeths:
https://github.com/longhorn/longhorn/issues/3826

image: wangsiye/longhorn-ui:3134d8b

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,6 +10,7 @@ import backup from './models/backup'
 import snapshot from './models/snapshot'
 import recurringJob from './models/recurringJob'
 import instanceManager from './models/instanceManager'
+import orphanedData from './models/orphanedData'
 
 // import assets
 import './assets/iconfont/iconfont.eot'
@@ -36,6 +37,7 @@ app.model(backup)
 app.model(volume)
 app.model(recurringJob)
 app.model(instanceManager)
+app.model(orphanedData)
 
 // 3. Router
 app.router(routerConfig)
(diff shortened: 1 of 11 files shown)


innobead:
cc @derekbit

derekbit:
Awesome!
Can we add a filter by Type? Thank you.

derekbit:
Awesome! Thank you.

derekbit:
Suggest some changes
1. Rename `Disk` to `Disk Name`
2. Rename `Path` to `Disk Path`
3. Add a `Disk Path` filter
4. Add `Directory Name` field (resource.spec.parameters[“DataName”])

Thank you. :)

smallteeths:
Regarding point 5, as the fields will be different for different types, can we add a select component to the page to allow the user to switch between different types to display different types of tables Instead of displaying multiple tables.  @derekbit 

截屏2022-05-06 下午1 29 36

derekbit:
> Regarding point 5, as the fields will be different for different types, can we add a select component to the page to allow the user to switch between different types to display different types of tables Instead of displaying multiple tables. @derekbit
> 
> 截屏2022-05-06 下午1 29 36

Perfect! Thank you ;)

derekbit:
> Regarding point 5, as the fields will be different for different types, can we add a select component to the page to allow the user to switch between different types to display different types of tables Instead of displaying multiple tables. @derekbit
> 
> 截屏2022-05-06 下午1 29 36

Then we don't need the `Type` field.

derekbit (approved):
LGTM.

innobead (approved):
LGTM

PR #520, opened 2022-06-06
feat(backing image) Add a UUID column to the backing image list

smallteeths:
image: wangsiye/longhorn-ui:4401b00

https://github.com/longhorn/longhorn/issues/4052


innobead (approved):

innobead:
cc @shuo-wu

PR #522, opened 2022-07-25
Fix(dockerfile) Upgrade webpack version

smallteeths:
Signed-off-by: Siye Wang <siye@rancher.com>

https://github.com/longhorn/longhorn/issues/4283

PR #523, opened 2022-07-26
Run the UI container with user UID as a user

mantissahz:
Kubernete might check the UID exists in func `verifyRunAsNonRoot`
for `security_context` in rke2 enviroment.

longhorn/longhorn#4266

Signed-off-by: James Lu <james.lu@suse.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,7 @@ EXPOSE 8000
 ENV LONGHORN_MANAGER_IP http://localhost:9500
 ENV LONGHORN_UI_PORT 8000
 
-USER nginx
+# Use the uid of the default user (nginx) from the installed nginx package
+USER 499
 
 CMD ["/bin/bash", "-c", "mkdir -p /var/config/nginx/ && cp -r /etc/nginx/* /var/config/nginx/; envsubst '${LONGHORN_MANAGER_IP},${LONGHORN_UI_PORT}' < /etc/nginx/nginx.conf.template > /var/config/nginx/nginx.conf && nginx -c /var/config/nginx/nginx.conf -g 'daemon off;'"]


smallteeths (approved):
LGTM

Dockerfile
```diff
@@ -26,6 +26,7 @@ EXPOSE 8000
 ENV LONGHORN_MANAGER_IP http://localhost:9500
 ENV LONGHORN_UI_PORT 8000
 
-USER nginx
+RUN usermod -u 499 nginx
```
innobead, line 29: 0-999 is usually for system accounts. Suggest changing to 1000.

innobead (changes requested):

innobead:
@mantissahz is this based on https://github.com/longhorn/longhorn-ui/pull/522 first?

Dockerfile
mantissahz: There is something weird that only using UID 499 would not have the `Permission denied` problem.
```
nginx: [alert] could not open error log file: open() "/var/log/nginx/error.log" failed (13: Permission denied)
2022/07/26 04:54:46 [warn] 10#10: duplicate MIME type "text/html" in /var/config/nginx/nginx.conf:7
2022/07/26 04:54:46 [emerg] 10#10: open() "/var/log/nginx/access.log" failed (13: Permission denied)
```

mantissahz:
> is this based on https://github.com/longhorn/longhorn-ui/pull/522 first?
No, it is not based on this PR. I skipped some package errors in my environment to make the container image for testing UID issue only.

innobead:
> > is this based on #522 first?
> 
> No, it is not based on this PR. I skipped some package errors in my environment to make the container image for testing UID issue only.

Got it, but we still need https://github.com/longhorn/longhorn-ui/pull/522, right?

smallteeths:
> > > is this based on #522 first?
> > 
> > 
> > No, it is not based on this PR. I skipped some package errors in my environment to make the container image for testing UID issue only.
> 
> Got it, but we still need #522, right?
> 
> Can you base on #522 to see if the build and UI work fine? Thanks.

I am working on the bug that build longhorn-ui image error. This bug is not fixed by my current #522 pr. 
The pr should be merged after I fix the above bug.

Dockerfile
mantissahz: OK, changed UID to 1000 and re-change the owner /var/log/nginx directory.

innobead: This should be because you changed the original uid of the Nginx user after the Nginx has been installed and related artifacts (including some folders) are already authorized. 

Could you try to use the uid of the nginx user instead? No need to do any uid update actually. 
```
USER 499
```

Dockerfile
```diff
@@ -26,6 +26,7 @@ EXPOSE 8000
 ENV LONGHORN_MANAGER_IP http://localhost:9500
 ENV LONGHORN_UI_PORT 8000
 
-USER nginx
+RUN usermod -u 1000 nginx && chown -R nginx:nginx /var/log/nginx
```
innobead, line 29: `chown -R nginx:nginx /var/log/nginx` seems not right, because it will possibly leave some other permission leftover belonging to the original nginx uid.

Dockerfile
mantissahz: Yes, it works, but would it always be UID 499?

innobead: Hmm because we relay on the package from BCI repo, this is not what we can change.

Dockerfile
mantissahz: Removed, only change `User 499`

Dockerfile
innobead: Have u verified if after this change,  longhorn-hi will run well on a CIS hardened cluster?

mantissahz: Yes, only update 'USER 499', it runs well on CIS hardened cluster I built.

Dockerfile
```diff
@@ -26,6 +26,6 @@ EXPOSE 8000
 ENV LONGHORN_MANAGER_IP http://localhost:9500
 ENV LONGHORN_UI_PORT 8000
 
-USER nginx
+USER 499
```
innobead, line 29: Suggest having a comment like 
```
# Use the uid of the default user (nginx) from the installed nginx package
User 499
```

mantissahz: Added, thanks

innobead (approved):

innobead:
Let's merge this after the UI PR is ready.

PR #541, opened 2022-09-16
Check if node schedulable condition is set before trying to read it

Foritus:
One of my longhorn nodes is not currently schedulable due to kubernetes nonsense, and this has broken the whole UI as it throws an exception on page load due to one node being bad. Also a little formatting as those one-liners are getting out of control.

image

### DIFF
--- a/src/utils/filter.js
+++ b/src/utils/filter.js
@@ -16,8 +16,8 @@ export function faultedVolume(data) { return data.filter((item) => item.state ==
 
 const isSchedulable = (node) => node.conditions && node.conditions.Schedulable && node.conditions.Schedulable.status.toLowerCase() === 'true'
                                                 && node.conditions.Ready && node.conditions.Ready.status.toLowerCase() === 'true'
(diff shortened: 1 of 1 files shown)


smallteeths (approved):
LGTM

innobead (approved):
LGMT.

@smallteeths Do you think if we can improve this kind of tedious null check in the future by using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining instead?

innobead:
@weizhe0422 because @smallteeths is not around, please help create an issue and add a reference to this PR, and related backports as well. Thanks.

cc @longhorn/qa

smallteeths:
> LGMT.
> 
> @smallteeths Do you think if we can improve this kind of tedious null check in the future by using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining instead?

OK, Some browsers did not support the `?.` syntax well when the project was first established.  It should be all supported now, and I will improve this kind of tedious null check in the next version by using `?.`

PR #558, opened 2022-12-15
fix(date) Initial date return null

smallteeths:
https://github.com/longhorn/longhorn/issues/5047


innobead (approved):

innobead:
@smallteeths need to backport to 1.4.x and 1.3.x.

PR #559, opened 2022-12-15
fix(system restore) Add warning message when deleting the system restore

smallteeths:
https://github.com/longhorn/longhorn/issues/5065


innobead (approved):

innobead:
@smallteeths Need to backport to 1.4.x.

PR #564, opened 2022-12-16
feat(create volume) Change the description of the snapshots trim sele…

smallteeths:
…ction ignore
https://github.com/longhorn/longhorn/issues/5080


innobead (approved):

innobead:
@smallteeths as I mentioned, there are 4 settings having an `ignored` value which means using the global setting, so we need to update all of them. 

image

innobead (approved):

innobead:
@Mergifyio backport v1.4.x

PR #572, opened 2022-12-22
[v1.4.x] fix(volume) Fix the Snapshot trim wrong label

smallteeths:
https://github.com/longhorn/longhorn/issues/5127


innobead:
ref: https://github.com/longhorn/longhorn-ui/pull/577

PR #574, opened 2022-12-22
[v1.4.x] fix(volume) Set the Snapshot Data Integrity init value to ignored

smallteeths:
https://github.com/longhorn/longhorn/issues/5126


innobead:
Use https://github.com/longhorn/longhorn-ui/pull/576 instead.

PR #591, opened 2023-02-09
feat(setting) Convert * to a multiplication character that the browse…

smallteeths:
https://github.com/longhorn/longhorn/issues/5338

### DIFF
--- a/src/routes/setting/setting.js
+++ b/src/routes/setting/setting.js
@@ -81,7 +81,11 @@ const form = ({
         break
     }
     let deprecatedSettings = setting.definition && setting.definition.type === 'deprecated'
-
+    // Convert * to a multiplication character that the browser can recognise
+    if (setting?.definition?.description) {
+      let reg = /\x20\*\x20/ig
+      setting.definition.description = setting.definition.description.replaceAll(reg, ' &times; ')
+    }
     return (
       <FormItem key={setting.id} className={'settings-container'} style={{ display: deprecatedSettings ? 'none' : 'block' }}>
         <span className={setting.definition.required ? 'ant-form-item-required' : ''} style={{ fontSize: '14px', fontWeight: 700, marginRight: '10px' }}>{setting.definition.displayName}{valuePropName === 'checked' ? ':' : ''}</span>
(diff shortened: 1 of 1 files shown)


src/routes/setting/setting.js
```diff
@@ -81,7 +81,10 @@ const form = ({
         break
     }
     let deprecatedSettings = setting.definition && setting.definition.type === 'deprecated'
-
+    // Convert * to a multiplication character that the browser can recognise
+    if (setting.id === 'guaranteed-engine-manager-cpu' && setting?.definition?.description) {
+      setting.definition.description = setting.definition.description.replaceAll('*', '&times;')
```
innobead, line 86: Can we make this general instead not just for guaranteed CPU setting?

innobead (changes requested):

src/routes/setting/setting.js
smallteeths: I'm not sure if the other settings just want to use `*` to say something important.

innobead: It should be fine, all characters in each description should be raw, so it's safe to escape *.

smallteeths: 截屏2023-02-09 下午12 54 19

An aggravated treatment such as the one shown above requires the `**~**` format. If &times;  is replaced globally it will not meet our expectations.

smallteeths: The pr still having problems , the `Guaranteed Replica Manager CPU` setting also needs to be replaced, I'm trying to think of a better way to do this

innobead: So that means we just regard each description as markdown-interpretable :thinking: 

Then, need to think about how to escape it from the raw string instead. `\*` will work?

smallteeths: React mark-down not recognized `\*`, `\*` is not work. I thought of a way to match /\x20\*\x20/ig based on the reg, because if it were a normal format expression it would all be *xxx*, **xx**. The /\x20\*\x20/ig reg will match * itself, so global substitution will do what we want. I'll provide a demo environment for you to check in a moment.

PR #606, opened 2023-03-28
fix(instance manager list) Fix the instancer manager name not wrappin…

smallteeths:
https://github.com/longhorn/longhorn/issues/5622

### DIFF
--- a/src/routes/instanceManager/InstanceManagerList.js
+++ b/src/routes/instanceManager/InstanceManagerList.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { Table, Button } from 'antd'
+import { Table } from 'antd'
 import { pagination } from '../../utils/page'
 import { sortTable } from '../../utils/sort'
 function list({ loading, dataSource, height, showRefCountVolumeModal }) {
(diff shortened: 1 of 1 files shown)


src/routes/instanceManager/InstanceManagerList.js
```diff
@@ -13,11 +13,9 @@ function list({ loading, dataSource, height, showRefCountVolumeModal }) {
...
       render: (text, record) => {
         return (
-          <div>
-            <Button type="link" onClick={() => showRefCountVolumeModal(record.volume)}>
-              {text}
-            </Button>
-          </div>
+          <a className="ant-btn-link" onClick={() => showRefCountVolumeModal(record.volume)}>
+            {text}
+          </a>
```
innobead, line 18: Can you explain why this change can solve the wrapping issue?

smallteeths: The property "white-space: nowrap;" on the class ".ant-btn" of antd `Button` prevents text from wrapping according to its width.

smallteeths: The main reason why Button does not wrap is the CSS property "white-space: nowrap;", which is added to the Button component by default in antd. To fix this issue while keeping the style unchanged, all we need to do is replace the Button component with an "a" link and add the ".ant-btn-link" class name.

innobead: :+1:

innobead (approved):

PR #611, opened 2023-04-21
Sync `package-lock.json` and use `npm ci`

votdev:
- Sync `package-lock.json`with `package.json` so that git does not constantly complain that the file has been modified after running `npm install`. This will allow the use of `npm ci` instead of `npm install`
- Use `npm ci` which install the NPM package dependencies AS specified. This is better for continuous integration.
- Update the README and Dockerfile


smallteeths (approved):
LGTM

PR #620, opened 2023-05-19
 feat(pvc) Add storage class name for pvc

smallteeths:
https://github.com/longhorn/longhorn/issues/3506


innobead (approved):
LGTM

ChanYiLin:
Hi @innobead @smallteeths 
sorry for not pay attention on the test
I just found a bug now
The storageClassname param should be passed to `"PV"Create` request instead of `"PVC"Create` request
Thanks

PR #621, opened 2023-05-26
fix(PV) Add storage class for PV

smallteeths:
https://github.com/longhorn/longhorn/issues/3506

image: wangsiye/longhorn-ui:79b50f6

### DIFF
--- a/src/models/volume.js
+++ b/src/models/volume.js
@@ -363,12 +363,12 @@ export default {
           pvAction.push(item)
         }
       })
-      yield pvAction.map(item => call(createVolumePV, { pvName: item.name, fsType }, item.actions.pvCreate))
+      const storageClassName = payload.params.storageClassName
+      yield pvAction.map(item => call(createVolumePV, { pvName: item.name, fsType, storageClassName }, item.actions.pvCreate))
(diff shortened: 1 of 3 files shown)


ChanYiLin:
Hi I have tested it, and there is one thing need to adjust

When I created PV/PVC from a restored Volume
I chose "use previous PVC" and left `StorageClassName` empty
The `StorageClassName` in payload send to backend would be `default`
but **it should be "" empty** string for backend to fill in the default longhorn storageClassName in setting.

Thanks!

smallteeths:
> Hi I have tested it, and there is one thing need to adjust
> 
> When I created PV/PVC from a restored Volume I left `StorageClassName` empty The `StorageClassName` in payload send to backend would be `default` but **it should be "" empty** string for backend to fill in the default longhorn storageClassName in setting.
> 
> Thanks!

done 
new image: wangsiye/longhorn-ui:79b50f6

ChanYiLin:
Hi @smallteeths 
I have tested it, LGTM, thanks!
cc @innobead

innobead (approved):
LGTM.

Thanks Jack and Siye.

PR #623, opened 2023-05-31
feat(node) Support create block disk type

smallteeths:
https://github.com/longhorn/longhorn/issues/5846

image: wangsiye/longhorn-ui:07f1d94


derekbit (approved):
LGTM

PR #626, opened 2023-06-05
fix(volume) Fix the bug of bluk attach request

smallteeths:
https://github.com/longhorn/longhorn/issues/6004


innobead:
@mergify backport v1.5.x

PR #643, opened 2023-08-01
fix: permission denied when running ui image

mantissahz:
It has a permission denied when running UI image to create tempfs directory, copy nginx config files to tempfs directory and open pid file in /run.
But it could be deployed on k8s cluster.

Ref: longhorn/longhorn#6430

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,8 @@ EXPOSE 8000
 ENV LONGHORN_MANAGER_IP http://localhost:9500
 ENV LONGHORN_UI_PORT 8000
 
+RUN mkdir -p /var/config/ && touch /var/run/nginx.pid && chown -R 499 /var/config /var/run/nginx.pid
+
 # Use the uid of the default user (nginx) from the installed nginx package
 USER 499
 


Dockerfile
```diff
@@ -26,6 +26,8 @@ EXPOSE 8000
 ENV LONGHORN_MANAGER_IP http://localhost:9500
 ENV LONGHORN_UI_PORT 8000
 
+RUN mkdir -p /var/config/ && touch /run/nginx.pid && chown -R 499 /var/config /run/nginx.pid
```
innobead, line 29: Why we need /run/nginx.pid? should it be in /var/run?

innobead (changes requested):

Dockerfile
mantissahz: Because it would show `[emerg] open() "/run/nginx.pid" failed (13: Permission denied)` without creating `/run/nginx.pid` and changing the owner. It will still happen when specifying `/var/run/nginx.pid`.
Specify the pid file location to `/var/config/nginx.pid` instead.

innobead: I see. Why I was asking is if you install nginx package, you will notice the default path for pid is /var/run/ngnix.pid. 

```shell
af432ea928fc:/ # cat /etc/nginx/nginx.conf | grep -i pid
#pid        /var/run/nginx.pid;
```

What you can do is `mkdir -p /var/run` and change the owner instead of changing the path of pid at runtime.  

```shell
RUN mkdir -p /var/config/ && mkdir -p /var/run && chown -R 499 /var/{config,run}
```

mantissahz: After some testings, I found the default path of nginx pid file is `/run/nginx.pid` (although the config file shows `#pid /var/run/nginx.pid;`) and `/var/run` is a symbolic link to `/run` (I am not sure if we could remove `/var/run` and re-create it as a new directory to save nginx pid file)
So I modified it to create pid file first in `/var/run` and change its owner to `nginx` (499)

innobead (approved):

PR #649, opened 2023-09-06
Object Storage (LEP 20230430)

m-ildefons:
User interface for managing the object stores of Longhorn.
This adds a new tab to the Longhorn UI, which is dedicated to managing object stores. It can list, create and delete object stores as well as show their operational status.
Each object store corresponds to an `ObjectStore` CRD in the K8s cluster, the Longhorn Manager implements a corresponding controller which orchestrates an [s3gw](https://github.com/aquarist-labs/s3gw) instance to implement that object store.

Related: https://github.com/longhorn/longhorn-manager/pull/2136
Related: https://github.com/longhorn/longhorn/pull/6640

LEP: https://github.com/longhorn/longhorn/pull/5832


src/models/objectStore.js
```diff
@@ -0,0 +1,86 @@
...
+            payload: location.pathname.startsWith('/objectstore') ? queryString.parse(location.search) : {},
+          })
+        }
+      })
+    },
+  },
+  effects: {
+    *query({ payload }, { call, put }) {
+      const data = yield call(listObjectStores, payload)
+      yield put({ type: 'listObjectStores', payload: { ...data } })
```
votdev, line 30: Is a `put` used by reason here? I assume it should be `get` instead.

votdev (changes requested):

src/models/objectStore.js
m-ildefons: It should be a `put`. `get` would cause errors

votdev (approved):

PhanLe1010:
Hey, in the current version, the access key is keep changing when user are typing the ObjectStore name in UI. I am guessing it is hashing the volume of the ObjectStore in realtime. This is bit strange as other people pointed out. Can we just generate a random string and stick with it?

PhanLe1010:
I couldn't figureout why we cannot access the s3gw-ui using your latest version at https://github.com/m-ildefons/s3gw-ui/tree/wip/longhorn-integration. Can you help?

Screenshot from 2023-11-20 17-03-18

m-ildefons:
This is odd. I've never seen that error, maybe @votdev has an idea where it comes from. Can you provide a reproducer or some kind of hint what your configuration looks like?

PR #656, opened 2023-10-30
Duplicate MIME type "text/html" in /var/config/nginx/nginx.conf

votdev:
According to the nginx docs (http://nginx.org/en/docs/http/ngx_http_gzip_module.html#gzip_types) the MIME type `text/html` does not need to be mentioned.

Fixes: https://github.com/longhorn/longhorn/issues/7002


innobead (approved):

innobead:
@mergify backport v1.5.x v1.4.x

PR #660, opened 2023-11-03
All components handle window resizing events incorrectly

votdev:
Fixes: https://github.com/longhorn/longhorn/issues/7036


innobead (approved):
LGTM

PR #270, opened 2020-05-16
fix(node status) change node status to reflect that it has been cordoned

smallteeths:
https://github.com/longhorn/longhorn/issues/1287

image: wangsiye/longhorn-ui:941d022

Signed-off-by: Siye Wang <siye@rancher.com>

PR #663, opened 2023-11-13
Display AutoEvicting on Node tab while auto evicting

ejweber:
longhorn/longhorn#2238

I originally considered also making AutoEvicting be a separate status from Unschedulable on the Dashboard, but this was too granular of a distinction in my opinion (and the opinion of a couple others I talked to). A user will probably only have one (or a small number) of auto evicting nodes at a time, and the Node tab is a good place to quickly identify it/them.

2238 Demo

### DIFF
--- a/src/routes/dashboard/components/resourceOverview.js
+++ b/src/routes/dashboard/components/resourceOverview.js
@@ -131,6 +131,7 @@ class ResourceOverview extends React.Component {
       faulted: faultedVolume(volume.data).length,
     }
 
+    // We do not represent autoEvicting in the resource overview. autoEvicting nodes are also unschedulable.
     const nodeInfo = {
       // Total. The total number of nodes.
       total: host.data.length,
@@ -238,6 +239,7 @@ class ResourceOverview extends React.Component {
       },
     }
 
+    // We do not represent autoEvicting in the resource overview. autoEvicting nodes are also unschedulable.
     const nodeInfoColors = [nodeStatusColorMap.schedulable.color, nodeStatusColorMap.unschedulable.color, nodeStatusColorMap.down.color, nodeStatusColorMap.disabled.color]
     const nodeInfoData = [
       { key: 'schedulable', name: 'Schedulable', value: nodeInfo.schedulable },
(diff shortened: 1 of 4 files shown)


ejweber:
As part of this PR, I reworked a few comments that I previously found confusing.

votdev:
Testing is currently blocked because of conflicts in https://github.com/longhorn/longhorn-manager/pull/2138.

ejweber:
> Conflicts are resolved for now, but I think it is fine to wait until those changes are actually in master to review this one. I can ping you when it is appropriate @votdev.

ejweber:
This one should be testable with components off of the master branches now @votdev.

m-ildefons:
I attempted to test this, by deploying a build of the current `master` branch of the longhorn manager using the `master` branch of the chart. However I found that the drained node is displayed as "unschedulable" rather than "auto evicting" both during eviction and after.
In addition to that, I noticed that the display color of the "unschedulable" status is grey for the disk and yellow for the node.
Am I possibly doing something wrong during testing this?

https://github.com/longhorn/longhorn-ui/assets/17141774/efe4f331-96a5-4f70-8ea4-afd022bb7c27

ejweber:
It took me a minute, but I recognized the problem, @m-ildefons. You don't actually have any replicas scheduled to your nodes in the attached video. We only display AutoEvicting as long as we are actively evicting replicas. Once we are done, we revert back to Unschedulable.

See https://github.com/longhorn/longhorn/blob/6ad473c1b68da13da8337489eefcc5cab26dd891/enhancements/20230905-automatically-evict-replicas-while-draining.md?plain=1#L203-L221 for a complete test case that will demonstrate the new UI behavior.

ejweber:
> In addition to that, I noticed that the display color of the "unschedulable" status is grey for the disk and yellow for the node.

This is not a regression, though I'm not sure whether or not it was originally intended. In https://github.com/longhorn/longhorn-ui/pull/270/files, @smallteeths added https://github.com/longhorn/longhorn-ui/blob/3b154752549fd827f134728d7c914558b1a2bd15/src/routes/host/DiskList.js#L16-L18. While it sets the text to `Unschedulable` under these circumstances, it uses the `disabled` `className`, causing the gray text.

There is no discussion of this in the PR. I will go ahead and fix it to be yellow unless you disagree?

votdev (approved):
LGTM

Peek 2023-12-12 16-02

ejweber:
@longhorn/dev-ui 

We need to get this one merged if possible so we can transition https://github.com/longhorn/longhorn/issues/2238 to `Ready For Testing` for the upcoming `v1.6.0` release.

innobead (approved):
LGTM

ejweber:
@mergify backport v1.5.x

PR #664, opened 2023-12-04
fix: display volume conditions

m-ildefons:
Fix displaying volume scheduling conditions in volume list and volume details page.

ref: longhorn/longhorn#7241

### DIFF
--- a/src/routes/volume/VolumeActions.js
+++ b/src/routes/volume/VolumeActions.js
@@ -226,7 +226,7 @@ function actions({
     }
   })
 
-  availableActions.push({ key: 'expandVolume', name: 'Expand Volume', disabled: selected?.conditions?.scheduled?.status?.toLowerCase() === 'false' })
+  availableActions.push({ key: 'expandVolume', name: 'Expand Volume', disabled: selected?.conditions?.Scheduled?.status?.toLowerCase() === 'false' })
   if (selected.controllers && selected.controllers[0] && !selected.controllers[0].isExpanding && selected.controllers[0].size !== 0 && selected.controllers[0].size !== selected.size && selected.controllers[0].size !== '0') {
     availableActions.push({ key: 'cancelExpansion', name: 'Cancel Expansion', disabled: false })
   }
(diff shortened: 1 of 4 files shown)


innobead:
> Fix displaying volume scheduling conditions in volume list and volume details page.
> 
> Fixes: [longhorn/longhorn#7241](https://github.com/longhorn/longhorn/issues/7241)

Let's not use special terms (ex: fix(s), close, etc) to automatically close the source issue after the PR gets merged except for some untestable changes. QA needs to verify it first before closing the issue.

ejweber:
Good find @m-ildefons!

Reviewing these initial changes, it becomes clear that https://github.com/longhorn/longhorn-manager/pull/2146 and its backports broke this aspect of the UI by converting conditions from lower case to camel case.

There was a discussion in the review (https://github.com/longhorn/longhorn-manager/pull/2146#discussion_r1328783013) on whether the change would break end-to-end testing, but we missed the impact on the UI.

cc @longhorn/qa

ejweber (changes requested):
This fixes part of the issue. We have the expected scheduling failure message back. However, we still don't have the Volume Conditions represented on the Volume Details page.

image

I think we should search for all lowercase uses of the four volume conditions from https://github.com/longhorn/longhorn-manager/blob/16c09fdd296bdd893ceaaa2717456d8ea8cc7036/k8s/pkg/apis/longhorn/v1beta2/volume.go#L101-L106 in the UI and modify them.

m-ildefons:
Damn, I missed those.
I did a quick `grep -r` for the three conditions, but I'm not sure this fix is quite right. In the end, this happens because we don't keep the json serialization and deserialization of the objects that we push across the LH API in sync between the manager and the UI.

ejweber (approved):
LGTM.

image

image

image

I do not like the way the `TooManySnapshots` condition shows up on the `Volume Details` page as just an icon (with no text), but I looked back at the original QA testing for the addition of this condition, and it confirms that this is the way it was intended.

https://github.com/longhorn/longhorn/issues/496#issuecomment-720658923

innobead (approved):

innobead:
The original change was introduced to 1.4 and 1.5. We need to backport this.

@mergify backport v1.5.x v1.4.x

innobead:
@mergify backport v1.5.x v1.4.x

PR #674, opened 2024-01-01
volume: validate volume creation according to the enabled data engines

derekbit:
https://github.com/longhorn/longhorn/issues/7505


innobead (approved):
LGTM

PR #680, opened 2024-01-15
fix(engineimage): use `EngineImage.status.Incompatible`

mantissahz:
Check the EngineImage Status with `.status.incompatible` instead of using `EngineImage.state` directly.

Ref: longhorn/longhorn#7683, longhorn/longhorn#7539

### DIFF
--- a/src/routes/engineimage/EngineImageList.js
+++ b/src/routes/engineimage/EngineImageList.js
@@ -31,10 +31,10 @@ function list({ loading, dataSource, deleteEngineImage }) {
       dataIndex: 'state',
       key: 'state',
       width: 100,
-      render: (text) => {
+      render: (text, record) => {
         return (
-          <div className={classnames({ [text.toLowerCase()]: true, capitalize: true })}>
-            {text.hyphenToHump()}
+          <div className={classnames({ [record.incompatible ? 'incompatible' : text.toLowerCase()]: true, capitalize: true })}>
+            {record.incompatible ? 'Incompatible' : text.hyphenToHump()}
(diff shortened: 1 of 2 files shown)


innobead:
@mantissahz  The commit message is confusing. `status.incompatible` is newly introduced not deprecated.

mantissahz:
>The commit message is confusing. status.incompatible is newly introduced not deprecated.

I will modify it soon.

src/routes/engineimage/EngineImageList.js
```diff
@@ -31,10 +31,10 @@ function list({ loading, dataSource, deleteEngineImage }) {
       dataIndex: 'state',
       key: 'state',
       width: 100,
-      render: (text) => {
+      render: (text, record) => {
         return (
-          <div className={classnames({ [text.toLowerCase()]: true, capitalize: true })}>
-            {text.hyphenToHump()}
+          <div className={classnames({ [record.incompatible ? 'incompatible' : text.toLowerCase()]: true, capitalize: true })}>
```
innobead, line 36: Does this class really take effect? I didn't see any style class `incompatible` or `deploying` have been defined, so I assume it did not work initially.

However, it's fine for now, since it did not have a right style before.

innobead (approved):

innobead:
@mergify backport v1.6.x

src/routes/engineimage/EngineImageList.js
innobead: cc @scures

PR #682, opened 2024-01-17
Longhorn snapshot space management

scures:
Ref: https://github.com/longhorn/longhorn/issues/7522

Enables global setting, volume creation, and volume update fields for maxCount and maxSize

- http://localhost:8080/#/setting should have the `Snapshot Maximum Count` field 
image
- http://localhost:8080/#/volume Creating a new volume should present you with the values for count & size
image
-  http://localhost:8080/#/volume active volumes should have the option to update the values for count & size
image
image
- Values shown as part of the volume details
image


FrankYang0529:
Hi @scures, thanks for the great work to build snapshot space management UI. Overall LGTM. I have few questions would like to check:

1. May we align the field name?
Screenshot 2024-01-18 at 11 52 05 AM

2. In UI, default snapshot max count for each volume should be 0, so webhook server has change to replace it with global snapshot max count setting.
Screenshot 2024-01-18 at 11 56 14 AM

3. Update Snapshot Max Count & Update Snapshot Max Size buttons can't work in the single volume page.

https://github.com/longhorn/longhorn-ui/assets/4927361/49d15bf9-6619-455b-8059-e38467e130db

innobead (approved):
LGTM.

innobead:
@mergify backport v1.6.x

innobead:
@scures Please move your ticket to ready-for-testing, so QA can help verify it.

FrankYang0529:
@scures it looks like there is an issue if snapshot max count is 0. The error message is from UI. I can create a volume with following YAML.

```yaml
apiVersion: longhorn.io/v1beta2
kind: Volume
metadata:
  name: test-vol-with-max-count
  namespace: longhorn-system
spec:
  frontend: blockdev
  numberOfReplicas: 2
  size: "2147483648"
  snapshotDataIntegrity: ignored
  snapshotMaxCount: 0
  snapshotMaxSize: "0"
```

Screenshot 2024-01-18 at 6 30 55 PM

innobead:
After discussing with @FrankYang0529 , we should let the backend handle that validation and mutation, so if the user's input is 0, it will be mutated with the global setting. 

@scures Please help with this change.

PR #684, opened 2024-01-18
Adds better messages for the tooltip

scures:
Ref: https://github.com/longhorn/longhorn/issues/7522

Changes: 
- Improves the tooltip message for `maxCount` so it's clear that 0 means global settings. Fixes the validation so it allows setting 0 back.
- Adds a tooltip for `maxSize` that lets the user know that the size should be at least twice as big as the volume.

image
image

### DIFF
--- a/src/routes/volume/UpdateSnapshotMaxCountModal.js
+++ b/src/routes/volume/UpdateSnapshotMaxCountModal.js
@@ -57,9 +57,21 @@ const modal = ({
               },
               {
                 validator: (rule, value, callback) => {
-                  if (value < 2 || value > 250 || !/^\d+$/.test(value)) {
+                  if (value === '' || typeof value !== 'number') {
+                    callback()
+                    return
+                  }
+                  if (value === 0) {
+                    callback()
+                  }
+                  if (value === 1) {
                     callback(
-                      'The value should be an integer between 2 and 250'
+                      'Set 0 to inherit global settings, or a value from 2 to 250'
+                    )
+                  }
+                  if (value < 2 || value > 250) {
+                    callback(
+                      'Set 0 to inherit global settings, or a value from 2 to 250'
                     )
                   } else {
                     callback()
--- a/src/routes/volume/UpdateSnapshotMaxCountModal.js
+++ b/src/routes/volume/UpdateSnapshotMaxCountModal.js
@@ -55,29 +55,6 @@ const modal = ({
                 required: true,
                 message: 'Please input volume snapshotMaxCount',
               },
-              {
-                validator: (rule, value, callback) => {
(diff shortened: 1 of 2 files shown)


src/routes/volume/UpdateSnapshotMaxCountModal.js
```diff
@@ -57,9 +57,21 @@ const modal = ({
...
+                  if (value === 1) {
+                    callback(
+                      'Set 0 to inherit global settings, or a value from 2 to 250'
+                    )
+                  }
+                  if (value < 2 || value > 250) {
                     callback(
-                      'The value should be an integer between 2 and 250'
+                      'Set 0 to inherit global settings, or a value from 2 to 250'
                     )
```
innobead, line 75: Don't think we need to have validation here, as we already have the same logic at the backend. Suggest removing them. Please do some tests after removing the validation. 

[/webhook/resources/volume/validator.go#L88-L94](https://github.com/longhorn/longhorn-manager/blob/aaeae7e46b938238c1f7cd80936f626d7ce073dc/webhook/resources/volume/validator.go#L88-L94)
```go
	if err := types.ValidateReplicaZoneSoftAntiAffinity(volume.Spec.ReplicaZoneSoftAntiAffinity); err != nil {
		return werror.NewInvalidError(err.Error(), "")
	}

	if err := types.ValidateReplicaDiskSoftAntiAffinity(volume.Spec.ReplicaDiskSoftAntiAffinity); err != nil {
		return werror.NewInvalidError(err.Error(), "")
	}

```

innobead: No, I meant both, but know your purpose here, so I am fine with that. The goal is no matter what ways users operate the resource (in this case, its volume) using kubectl or UI, the behavior/result should be identical.

After removing the validation in the modal, when updating that value in the modal, will users still have a validation before sending the request to the backend? or it will just rely on validation at backend?

innobead: Thanks for the update. I would suggest having backend handle the validation for both cases. Later, if having any change, we can only fix one place.

At front end, we can just focus basic validation like type.

If we will have a better validation mechanism in the future such as sharing the same validation as backend, then we can have frontend validation for better UX. WDYT?

FrankYang0529 (approved):
LGTM. Thank you!

src/routes/volume/UpdateSnapshotMaxCountModal.js
innobead: @scures Please handle this part later. I will merge this PR for RC2 first.

innobead:
@mergify backport v1.6.x

PR #688, opened 2024-01-22
deals with floating point

scures:
Longhorn Ref: https://github.com/longhorn/longhorn/issues/7522

- Fixes an issue where having 2.2Gi would be displayed as `2.2000000001862645`
image


FrankYang0529:
@scures may we also change the term `Max Count` to `Max Size` in Update Snapshot Max Size form? Thank you.

FrankYang0529 (approved):
LGTM. Thank you.

PR #696, opened 2024-02-13
Strange placeholder behaviour in input field

votdev:
Replace `it’s` with `it is` because the apostrophe is rendered with a big space with the configured `AvenirNext-Regular` font-family. This is the only occurrence of `it’s` in the whole UI, everywhere else `it is` is used. Perhaps for certain reasons.

#### Which issue(s) this PR fixes:
https://github.com/longhorn/longhorn/issues/7917

### DIFF
--- a/src/routes/backingImage/CreateBackingImage.js
+++ b/src/routes/backingImage/CreateBackingImage.js
@@ -209,7 +209,7 @@ const modal = ({
                 required: false,
               },
             ],
-          })(<Input placeholder="Ask Longhorn to validate the SHA512 checksum if it’s specified here." />)}
+          })(<Input placeholder="Ask Longhorn to validate the SHA512 checksum if it is specified here." />)}
         </FormItem>
       </Form>
     </ModalBlur>


votdev:
> It seems to be an env situation with the font. I see that the list of fonts goes for a bit and not sure if this is Antd configuration but we might be able to re-write it and use something more generic if Avenir is missing.
> 
> `AvenirNext-Regular, "Helvetica Neue", "lucida grande", PingFangHK-Light, STHeiti, "Heiti SC", "Hiragino Sans GB", "Microsoft JhengHei", "Microsoft Yahei", SimHei, "WenQuanYi Micro Hei", "Droid Sans", Roboto, Helvetica, Tahoma, Arial, "sans-serif"`
> 
> It renders Ok on MacOS image

Adapting the font might be another solution. But as mentioned above, i've not found any other use of an apostrophe in the messages. There is also another cosmetic thing, i've never seen the use of `’` as apostrophe; i think `'` is more common. So i vote for removing it.

a110605:
Test on the latest Chrome, firefox and edge can't reproduce this issue. Looks like the browser font issue.
But it's okay to minor replace `it's` to `it is`.
 
@votdev , could you rebase with master ?

a110605 (approved):
LGTM

PR #724, opened 2024-05-09
Correct original empty string won't trigger unsaved modal

a110605:
Fix defect found by QA: https://github.com/longhorn/longhorn/issues/7497#issuecomment-2102212463

if `targetSettingOldValue` is empty string or zero, the original logic won't record the unsaved change.

After fix

https://github.com/longhorn/longhorn-ui/assets/5744158/8e1449c8-2618-4f86-998e-dc267000c45e


torchiaf (approved):
LGTM

a110605:
Verified all settings change locally are good to trigger unsaved modal.

PR #729, opened 2024-05-20
Allow download mulitple backing images with one click and move clean up action in detail modal

a110605:
## Goal 
 - Allow user to sort the backing image list
 - Add bulk download button to allow user download mutliple backing images at once
 - Remove disk cleanup action in operation dropdown and move into detail modal 

## Issue
https://github.com/longhorn/longhorn/issues/7293
 

## Result
Sort the backing image list
**sort**

Bulk download
d

Move cleanup action in detail modal
clean

### DIFF
--- a/src/services/backingImage.js
+++ b/src/services/backingImage.js
@@ -28,9 +28,21 @@ export async function deleteBackingImage(params) {
   }
 }
 
+export async function bulkDownload(images) {
+  for (const image of images) {
+    if (image?.name) {
+      // eslint-disable-next-line no-underscore-dangle
+      const downloadUrl = `${window.__pathname_prefix__}${window.__pathname_prefix__.endsWith('/') ? '' : '/'}v1/backingimages/${image.name}/download`
+      // specific way to download multiple files with one click, User may allow the browser to download multiple files at once
+      Object.assign(document.createElement('a'), { href: downloadUrl, download: image.name }).click()
+    }
+  }
+}
+
 export async function download(params) {
   if (params && params.name) {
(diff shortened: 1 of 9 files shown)


src/services/backingImage.js
```diff
@@ -28,9 +28,21 @@ export async function deleteBackingImage(params) {
   }
 }
 
+export async function bulkDownload(images) {
+  for (const image of images) {
+    if (image?.name) {
+      // eslint-disable-next-line no-underscore-dangle
+      const downloadUrl = `${window.__pathname_prefix__}${window.__pathname_prefix__.endsWith('/') ? '' : '/'}v1/backingimages/${image.name}/download`
+      // specific way to download multiple files with one click, User may allow the browser to download multiple files at once
+      Object.assign(document.createElement('a'), { href: downloadUrl, download: image.name }).click()
```
a110605, line 37: Current backing image [download API](https://github.com/a110605/longhorn-ui/blob/2021fa07e0b88e4ee36debb1badbd01ef28fa31d/src/services/backingImage.js#L42) only support download one image at a time
```
v1/backingimages/${params.name}/download
```
This workaround is to create multiple download link and clikc, but need user to give browser permission to do that. So I add a note in modal to let user know
Screenshot 2024-05-18 at 4 14 46 PM
Screenshot 2024-05-20 at 10 20 41 AM




The best approach is to zip the selected image from backend and has a new API to download that zip file. This is also how dropbox and google drive support download multiple files.


Refer :  https://stackoverflow.com/questions/52072098/how-to-download-multiple-files-with-window-open#comment91093930_52072098:~:text=Better-,to,-return%20a%20ZIP

torchiaf (approved):
LGTM

ChanYiLin:
LGTM, thanks!

PR #736, opened 2024-05-29
Filter out non-ready status backing images in volume create modal dropdown

a110605:
## Goal
- Filter out non-ready status backing image in volume create modal dropdown
- Add more backing images disk state colors

## Issue 
Implement extra improvement based on discussion [[IMPROVEMENT] BackingImage UI improvement](https://github.com/longhorn/longhorn/issues/7293#issuecomment-2134688954)
 

## Result

Screenshot 2024-05-29 at 12 38 27 PM

Only let user choose `READY` status backing images

Screenshot 2024-05-29 at 12 38 57 PM

### DIFF
--- a/src/routes/volume/index.js
+++ b/src/routes/volume/index.js
@@ -59,6 +59,7 @@ import {
 } from './helper'
 import { healthyVolume, inProgressVolume, degradedVolume, detachedVolume, faultedVolume, filterVolume, isVolumeImageUpgradable, isVolumeSchedule } from '../../utils/filter'
 import C from '../../utils/constants'
+import { hasReadyBackingDisk } from '../../utils/status'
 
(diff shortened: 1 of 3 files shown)


src/routes/volume/index.js
```diff
@@ -761,7 +762,7 @@ class Volume extends React.Component {
       v1DataEngineEnabled,
       v2DataEngineEnabled,
       diskTags,
-      backingImages,
+      backingImages: backingImages?.filter(image => hasReadyBackingDisk(image)) || [],
```
a110605, line 765: I've consulted @ChanYiLin if there exists at least one `READY` status backing image disk, user is able to create volume from that backing image and attach to node.

image

torchiaf (approved):
LGTM

PR #752, opened 2024-06-17
build: update base image to bci 15.6

innobead:


innobead:
@mergify backport v1.6.x v1.5.x

PR #759, opened 2024-06-24
fix the changed value using bulk action won't update the config states

a110605:
Found a bug from previous implementation. Some changed value won't update to the config states. That user may restore or create wrong volume while using bulk action. (See below comment for root cause)

Screenshot 2024-06-24 at 4 48 25 PM

### Issue

[[IMPROVEMENT] Restore Latest Backup should be applied with BackingImage name value](https://github.com/longhorn/longhorn/issues/7560)

### Test Result

Bulk action for restore latest backup

https://github.com/longhorn/longhorn-ui/assets/5744158/425025b4-d2a2-4129-8d8f-430d10afb44b

Bulk action for create DR volume

https://github.com/longhorn/longhorn-ui/assets/5744158/d4560f4d-6173-43ee-94e8-c049ed79d1a5

### DIFF
--- a/src/routes/backup/BulkCreateStandbyVolumeModal.js
+++ b/src/routes/backup/BulkCreateStandbyVolumeModal.js
@@ -1,6 +1,6 @@
 import React, { useState } from 'react'
 import PropTypes from 'prop-types'
-import { Form, InputNumber, Select, message, Spin, Checkbox, Tooltip, Tabs, Button, Input } from 'antd'
+import { Form, InputNumber, Select, message, Spin, Checkbox, Tooltip, Tabs, Button, Input, Popover, Alert } from 'antd'
 import { ModalBlur } from '../../components'
 import { formatMib } from '../../utils/formatter'
 
@@ -26,6 +26,7 @@ const modal = ({
   nodeTags,
   diskTags,
   tagsLoading,
+  backupVolumes,
   backingImages,
   v1DataEngineEnabled,
   v2DataEngineEnabled,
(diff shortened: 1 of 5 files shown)


src/routes/backup/BulkCreateStandbyVolumeModal.js
```diff
@@ -114,30 +90,52 @@ const modal = ({
...
-      message.error('Please fill in all required fields before switching to another disaster recover volume tab', 5)
+      message.error('Please correct the error fields before switching to another tab', 5)
       return
     }
     validateFields((errors) => {
-      if (errors) {
-        return
-      }
-      const data = {
-        ...getFieldsValue(),
```
a110605, line 128: Previous implementation I choose to set current configuration tab in onChange / onSelect callback by using `getFieldsValue()`.

But that has problem not get the latest updated value in the action callback

torchiaf (approved):
I tested the new behavior and all is working as expected;
I have a couple of questions:

- The users have to change the default `Volume Name` value to create the restored volume. Before to that, a tooltip message is displayed.
Unless we are using the same approach in other forms, would it be better to show a validation error or warning instead ?

image

- The `Backing Image` field is correctly disabled; should it be filled with the source value ? In my case it's empty

image

Otherwise LGTM

a110605:
> I tested the new behavior and all is working as expected; I have a couple of questions:
> 
> * The users have to change the default `Volume Name` value to create the restored volume. Before to that, a tooltip message is displayed.
>   Unless we are using the same approach in other forms, would it be better to show a validation error or warning instead ?
> 
> image
> 
> * The `Backing Image` field is correctly disabled; should it be filled with the source value ? In my case it's empty
> 
> image
> 
> Otherwise LGTM

Thanks for testing.
1. The aim of tooltip is to let user know some tips. This is previous author implementation to show tooltip when editing the form. I think it's okay and follow this pattern. 
In fact, the scenario to `create DR volume` and `Restore latest backup` are  create the diaster backup volumes in other clusters. (See [LH doc](https://longhorn.io/docs/1.7.0/snapshots-and-backups/setup-disaster-recovery-volumes/))

4. Only the backup created from a volume created using backing image will prefilled this backing image field. Others will leave this filed empty. The reason to lock this field is choose other backing image when restoring backup will cause unexpected error.
Screenshot 2024-06-25 at 9 27 39 AM

PR #766, opened 2024-07-04
Support create full backup in recurring backup task job and create backup modal

a110605:
### What this PR does / why we need it
- [x] add Full Backup checkbox in create backup modal
- [x] add parameters with dropdown and numberInput fields in create recurring backup task
- [x] allow user edit parameters in recurring modal
- [x] add backup mode column in backup table
### Issue
[[FEATURE UI] Add parameters support to the Backup and RecurringJob](https://github.com/longhorn/longhorn/issues/8291)

### Test Result
Screenshot 2024-07-03 at 5 13 31 PM

Screenshot 2024-07-03 at 5 13 05 PM

**Recurring job**

https://github.com/longhorn/longhorn-ui/assets/5744158/d8faf075-3308-4d9d-943a-63ed0b4a243d

**Edit Recurring job (allow unselect parameters key)**
edit

### Additional documentation or context
Try this PR change using `a110605/longhorn-ui:0704issue8291amd64`

PR #768, opened 2024-07-08
fix: bring backupMode in volume/bulkBackup payload

a110605:
In previous [PR](https://github.com/longhorn/longhorn-ui/pull/766), I didn't notice there is bulk `Create Backup` action in volume page, we should bring `backupMode` in payload as well.

### Issue
https://github.com/longhorn/longhorn/issues/8947

### Test Result

Screenshot 2024-07-08 at 3 11 50 PM

Screenshot 2024-07-08 at 3 12 32 PM

### DIFF
--- a/src/models/volume.js
+++ b/src/models/volume.js
@@ -472,7 +472,15 @@ export default {
     *bulkBackup({
       payload,
     }, { put }) {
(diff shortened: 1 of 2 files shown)


derekbit (approved):
LGTM

PR #773, opened 2024-07-17
ignore case sensitive when search in instance manager image page

a110605:
The bug comes from currentState : "running" is lowercase. but user searches "Running"
 
Ignore case sensitive when search in instance manager image page

### Issue
https://github.com/longhorn/longhorn/issues/9010

### Test Result

https://github.com/user-attachments/assets/91005490-2f3a-4a5c-bd62-305a399d7bd6

Note. this PR need backport to v1.7.x once merged

### DIFF
--- a/src/models/instanceManager.js
+++ b/src/models/instanceManager.js
@@ -25,8 +25,10 @@ export default {
       payload,
     }, { call, put }) {
       const data = yield call(query, payload)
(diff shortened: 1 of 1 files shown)


derekbit (approved):

derekbit:
@mergify backport v1.7.x

PR #812, opened 2024-11-13
chore: replace babel-eslint with @babel/eslint-parser

houhoucoop:
Replaces `babel-eslint` with `@babel/eslint-parser`, which requires `ESLint 7.5.0` or higher.
The bot PR failed, so we’re upgrading ESLint manually in this PR.

### Issue
https://github.com/longhorn/longhorn-ui/pull/776

### Test Result
0 error from `npm run lint`

Some new rules are disabled to avoid many file changes in this PR. 
They can be reviewed again when we introduce [Prettier](https://prettier.io/) in the future.


.eslintrc
```diff
@@ -82,16 +79,21 @@
...
   },
-  "parser": "babel-eslint",
+  "parser": "@babel/eslint-parser",
   "parserOptions": {
     "sourceType": "module",
-    "ecmaVersion": 8,
+    "ecmaVersion": 2020,
     "ecmaFeatures": {
-      "jsx": true,
-      "experimentalObjectRestSpread": true
```
houhoucoop, line 93: removed in eslint v6.
ref: https://eslint.org/docs/latest/use/migrating-to-6.0.0#-the-deprecated-experimentalobjectrestspread-option-has-been-removed

package.json
```diff
@@ -78,12 +78,11 @@
     "css-loader": "^2.1.1",
     "cssnano": "^4.1.10",
     "es5-imcompatible-versions": "^0.1.34",
-    "eslint": "^5.16.0",
-    "eslint-config-airbnb": "^17.1.0",
-    "eslint-plugin-compat": "^2.2.0",
```
houhoucoop, line 83: I don't think this is used in the code.

a110605 (approved):
`npm run lint` is good.


Screenshot 2024-11-14 at 4 25 17 PM

PR #833, opened 2024-12-20
feat: add support for additional nginx config via /var/config/nginx/conf.d/*.conf

steled:

It would be great to be able to configure nginx via additional config files from the /var/config/nginx/conf.d directory.

## Summary by CodeRabbit

- **New Features**
	- Enhanced configuration management by allowing dynamic inclusion of multiple configuration files.

### DIFF
--- a/nginx.conf.template
+++ b/nginx.conf.template
@@ -45,5 +45,7 @@ http {
              root /web/dist;
         }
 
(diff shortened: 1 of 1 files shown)


steled:
closing as I missed correct commit message

PR #834, opened 2024-12-20
feat: add support for additional nginx configuration

steled:

It would be great to be able to configure nginx via additional config files from the /var/config/nginx/conf.d directory.

## Summary by CodeRabbit

- **New Features**
	- Enhanced configuration management by allowing dynamic inclusion of multiple configuration files in Nginx.

- **Bug Fixes**
	- No specific bug fixes were noted in this release.


houhoucoop:
I’m thinking of closing this PR since we might not need it, any ideas?

steled:
I thought it would bei good to have this possibility as I'm using cilium ingress and currently it is not possible to configure it with annotations so I need to configure them directly in the apps.

PR #896, opened 2025-03-18
fix: expand volume option is greyed under Volume tab but working in the volume detail section

houhoucoop:
- `Expand Volume` should be consistent across the volume details page and bulk actions
- `Expand Volume` should be enabled for detached volume

### Issue
[[BUG] Expand Volume option is greyed under Volume tab but working in the volume detail section.  #7529
](https://github.com/longhorn/longhorn/issues/7529)

### Test Result
- Go to the volume page
- Select a detached volume and click on bulk actions
- In the dropdown menu, the `Expand Volume` option should be enabled
- Navigate to the volume details page, and the `Expand Volume` should also be enabled there

Screenshot 2025-03-18 at 1 26 13 PM (2)
Screenshot 2025-03-18 at 1 31 47 PM (2)

N/A

## Summary by CodeRabbit

- **Bug Fixes**
  - Adjusted the conditions for the "Expand Volume" action to enable it when the volume is not actively scheduled, enhancing the consistency of the action’s availability and preventing unintended expansion based on the volume’s scheduling status.

### DIFF
--- a/src/routes/volume/VolumeActions.js
+++ b/src/routes/volume/VolumeActions.js
@@ -248,7 +248,7 @@ function actions({
   })
 
   availableActions.push({ key: 'cloneVolume', name: 'Clone Volume', disabled: selected.standby || isRestoring(selected) })
-  availableActions.push({ key: 'expandVolume', name: 'Expand Volume', disabled: selected?.conditions?.Scheduled?.status?.toLowerCase() === 'false' })
+  availableActions.push({ key: 'expandVolume', name: 'Expand Volume', disabled: selected?.conditions?.Scheduled?.status?.toLowerCase() !== 'true' })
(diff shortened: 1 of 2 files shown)


src/routes/volume/VolumeActions.js
```diff
@@ -248,7 +248,7 @@ function actions({
   })
 
   availableActions.push({ key: 'cloneVolume', name: 'Clone Volume', disabled: selected.standby || isRestoring(selected) })
-  availableActions.push({ key: 'expandVolume', name: 'Expand Volume', disabled: selected?.conditions?.Scheduled?.status?.toLowerCase() === 'false' })
+  availableActions.push({ key: 'expandVolume', name: 'Expand Volume', disabled: selected?.conditions?.Scheduled?.status?.toLowerCase() !== 'true' })
```
houhoucoop, line 251: Disable the button if the value is `false` or `undefined`.

a110605 (approved):

PR #901, opened 2025-04-14
fix: undefined error in filterRemoved()

a110605:
fix: undefined error in filterRemoved()

### Issue
https://github.com/longhorn/longhorn/issues/10529

### DIFF
--- a/src/models/snapshot.js
+++ b/src/models/snapshot.js
@@ -53,7 +53,7 @@ let filterRemoved = (data) => {
       if (parent !== '') {
         let parentEntity = data.find(el => el.name === parent)
 
-        if (parentEntity) {
+        if (parentEntity && Array.isArray(parentEntity.children)) {
           let pChildArray = parentEntity.children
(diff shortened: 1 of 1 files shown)


src/models/snapshot.js
```diff
@@ -53,12 +53,14 @@ let filterRemoved = (data) => {
...
-          .concat(item.children)
-        parentEntity.children = pChildArray
+        if (parentEntity) {
+          let pChildArray = parentEntity.children
+          let itemPos = pChildArray.findIndex(ele => ele === item.name)
+          pChildArray = pChildArray.slice(0, itemPos)
+            .concat(pChildArray.slice(itemPos + 1, pChildArray.length))
+            .concat(item.children)
+          parentEntity.children = pChildArray
+        }
```
houhoucoop, line 63: how about `if (parentEntity && Array.isArray(parentEntity.children))`?

a110605: Updated.

houhoucoop (approved):
LGTM

PR #903, opened 2025-04-22
feat: v2 volume supports UBLK frontend

houhoucoop:
- Add the `Frontend` options during volume creation based on the selected data engine version
- Display the `Frontend` setting `UBLK` on the volume details page

### Issue
[[TASK] [UI] [FEATURE] v2 volume supports UBLK frontend #10735](https://github.com/longhorn/longhorn/issues/10735#issuecomment-2817729136)

### Test Result
1. Ensure the `UBLK` kernel module is loaded on each node by running `modprobe ublk_drv`
2. Navigate to the volume page and click on `Create Volume`
3. In the popup modal:
   - For `v1` data engine, only the following `Frontend` options should be available:
     - `Block Device`
     - `iSCSI`
   - For `v2` data engine, the following `Frontend` options should be available:
     - `Block Device`
     - `NVMf `
     - `UBLK`
4. Select the `v2` data engine and choose `UBLK` as the frontend
5. Attach the volume, it should attach successfully
6. Go to the volume detail page, the `Frontend: UBLK` should be displayed in the `Volume Details` section

https://github.com/user-attachments/assets/f4e60b9e-1474-4b71-a496-560216d529a4

N/A

### DIFF
--- a/src/routes/volume/helper/index.js
+++ b/src/routes/volume/helper/index.js
@@ -711,8 +711,10 @@ export function needToWaitDone(state, replicas) {
 }
 
 export const frontends = [
-  { label: 'Block Device', value: 'blockdev' },
-  { label: 'iSCSI', value: 'iscsi' },
+  { label: 'Block Device', value: 'blockdev', dataEngine: ['v1', 'v2'] },
+  { label: 'iSCSI', value: 'iscsi', dataEngine: ['v1'] },
+  { label: 'NVMF', value: 'nvmf', dataEngine: ['v2'] },
+  { label: 'UBLK', value: 'ublk', dataEngine: ['v2'] },
 ]
 
 export function disabledSnapshotAction(volume, modelState) {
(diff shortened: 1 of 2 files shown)


src/routes/volume/helper/index.js
```diff
@@ -711,8 +711,10 @@ export function needToWaitDone(state, replicas) {
 }
 
 export const frontends = [
-  { label: 'Block Device', value: 'blockdev' },
-  { label: 'iSCSI', value: 'iscsi' },
+  { label: 'Block Device', value: 'blockdev', dataEngine: ['v1', 'v2'] },
+  { label: 'iSCSI', value: 'iscsi', dataEngine: ['v1'] },
+  { label: 'NVMF', value: 'nvmf', dataEngine: ['v2'] },
```
derekbit, line 716:
```suggestion
  { label: 'NVMf', value: 'nvmf', dataEngine: ['v2'] },
```

derekbit (approved):
LGTM

PR #906, opened 2025-04-29
feat: Add "Running Replicas" to volumes table

xelab04:

- Add "Running Replicas" column to the table of volumes.
- Avoids having to click on a volume to see how many replicas are present.
- Helps quickly check if the number of actual replicas is equal to the desired replica count.

### Issue

[[FEATURE] Running replicas field in volume table #10817](https://github.com/longhorn/longhorn/issues/10817)

### Test Result

- Shows number of replicas present in the cluster
image

### DIFF
--- a/src/models/volume.js
+++ b/src/models/volume.js
@@ -148,6 +148,10 @@ export default {
         data.data = data.data.filter(item => item.controller && item.controller.hostId
           && payload.keyword.split(',').indexOf(item.controller.hostId) > -1)
       }
+      // Add runningreplicas field to each volume
+      data.data.forEach(volume => {
+        volume.numberOfRunningReplicas = volume.replicas ? volume.replicas.length : 0
+      })
       sortVolume(data.data)
       yield put({ type: 'queryVolume', payload: { ...data } })
       yield put({ type: 'clearSelection' })
--- a/src/models/volume.js
+++ b/src/models/volume.js
@@ -148,12 +148,6 @@ export default {
         data.data = data.data.filter(item => item.controller && item.controller.hostId
           && payload.keyword.split(',').indexOf(item.controller.hostId) > -1)
       }
-      // Add runningreplicas field to each volume
-      data.data.forEach(volume => {
-        volume.numberOfRunningReplicas = Array.isArray(volume.replicas)
-          ? volume.replicas.filter(replica => replica.running === true).length
-          : 0
-      })
       sortVolume(data.data)
       yield put({ type: 'queryVolume', payload: { ...data } })
       yield put({ type: 'clearSelection' })
--- a/src/models/volume.js
+++ b/src/models/volume.js
@@ -150,7 +150,9 @@ export default {
       }
       // Add runningreplicas field to each volume
       data.data.forEach(volume => {
-        volume.numberOfRunningReplicas = volume.replicas ? volume.replicas.length : 0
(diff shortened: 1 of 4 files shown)


xelab04:
A few points of contention which I would be happy to change:

- Column name is "Running Replicas" which may be ambiguous as it counts all replicas present, not filtering for readiness.
- Field is only `width: 100` which makes the table header take two lines. I made this decision as it felt unnecessary to take up more space for just a number.

xelab04:
Since then, I have removed the custom column and integrated it directly with the existing replicas column. It now looks like this:
image

I have also added a "check" so the number of actual replicas (on the left) are those which Longhorn considers to be running. However, this means it also counts volumes which are being rebuilt as "running".

If you want to test, I already put a docker image out - `xelab04/longhorn-10817:v1.8.0-test`

src/routes/volume/VolumeList.js
```diff
@@ -462,7 +462,20 @@ function list({
...
+    },
+    {
+      title: 'Running Replicas',
+      dataIndex: 'runningreplicas',
+      key: 'runningreplicas',
+      width: 100,
+      render: (text, record) => {
+        return (
+          <div>
+            {record.numberOfRunningReplicas}
```
houhoucoop, line 478: Redundant code?

src/routes/volume/VolumeList.js
```diff
@@ -462,7 +462,20 @@ function list({
       render: (text, record) => {
         return (
           <div>
-            {record.numberOfReplicas}
+            {record.numberOfRunningReplicas} / {record.numberOfReplicas}
+          </div>
+        )
+      },
+    },
```
houhoucoop, line 469: [Nit] I suggest using `Replicas (Running / Desired)` as the column title for better clarity.

src/models/volume.js
```diff
@@ -148,6 +148,12 @@ export default {
         data.data = data.data.filter(item => item.controller && item.controller.hostId
           && payload.keyword.split(',').indexOf(item.controller.hostId) > -1)
       }
+      // Add runningreplicas field to each volume
+      data.data.forEach(volume => {
+        volume.numberOfRunningReplicas = Array.isArray(volume.replicas)
+          ? volume.replicas.filter(replica => replica.running === true).length
+          : 0
+      })
```
houhoucoop, line 156: `numberOfRunningReplicas` is only used for display in the table, I suggest handling it directly in the table component so we don’t need to manage it here and again in websocket updates.

xelab04: I've moved it to VolumeList.js so now it's handled in the table render segment. Should be better now.

src/routes/volume/VolumeList.js
xelab04: I did that change. Also increased the width of the column to cater for the extra width taken up.

derekbit (approved):
Technically, LGTM

xelab04:
> @xelab04 DCO failed, need to fix the commit sign-off.

Thank you! I had swapped to my work laptop and thought the name in the sign-off had to match my github username. Fixed it.

houhoucoop (approved):
LGTM, thanks for the contribution.

derekbit:
Thanks @xelab04.
Good cooperation @xelab04 @c3y1huang @houhoucoop!

derekbit:
@mergify backport v1.9.x

PR #947, opened 2025-08-25
feat: add UI Warning for Force-Detach Actions to Prevent Out-of-Sync Kubernetes and Longhorn VolumeAttachments

houhoucoop:
- Display a warning message in the detached confirmation modal

### Issue
[[UI][IMPROVEMENT] Add UI Warning for Force-Detach Actions to Prevent Out-of-Sync Kubernetes and Longhorn VolumeAttachments #9944](https://github.com/longhorn/longhorn/issues/9944)

### Test Result
- Navigate to volume or volume detail page, click `Detach` on the target volume
- If the volume attachment type is `csi-attacher`, verify that a warning message is displayed in the confirmation modal
Screenshot 2025-08-25 at 3 33 36 PM (2)

N/A

### DIFF
--- a/src/routes/volume/DetachHost.js
+++ b/src/routes/volume/DetachHost.js
@@ -65,11 +65,33 @@ const modal = ({
     onOk: handleOk,
   }
 
-  const description = (<div>
-    Detaching a Volume when it is being used by a running Kubernetes Pod will result in
-    crashing of the Pod and possible loss of data. The Volume cannot be used by the
-    Kubernetes again until the original Pod is deleted. Are you sure you want to detach the volume?
-  </div>)
+  const getAlertDescription = (item) => {
+    const va = item?.volumeAttachment
+    if (!va) return null
+
+    const attachments = Object.values(va.attachments || {})
+
+    if (attachments.some(att => att.attachmentType === 'csi-attacher')) {
+      return (
+        <div>
+          Force-detaching this volume may cause it to become inconsistent with Kubernetes VolumeAttachment resources.<br />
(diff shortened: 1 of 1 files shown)


src/routes/volume/DetachHost.js
```diff
@@ -65,11 +65,33 @@ const modal = ({
...
+    if (!va) return null
+
+    const attachments = Object.values(va.attachments || {})
+
+    if (attachments.some(att => att.attachmentType === 'csi-attacher')) {
+      return (
+        <div>
+          Force-detaching this volume may cause it to become inconsistent with Kubernetes VolumeAttachment resources.<br />
+          This can prevent workloads from attaching the volume correctly and may lead to Pod failures or potential data loss.
+        </div>
```
houhoucoop, line 79: Not sure if the warning message is clear or meaningful to users, any suggestions are welcome.

sushant-suse: We can also write something like this:
"
Forcing this volume to detach could break its connection to Kubernetes's VolumeAttachment resources. This could prevent your applications from attaching volumes correctly, potentially causing Pods to fail or leading to data loss.
"

houhoucoop:
> > If the volume attachment type is csi-attacher
> 
> Hello @houhoucoop, only show the message when the csi-attacher ticket is satisfied.

Do you mean when any of the csi-attacher tickets is satisfied? 
Can there be multiple Volume Attachments? e.g., if one csi-attacher ticket is satisfied and another is not, should the message still be shown?

sushant-suse (approved):

PR #962, opened 2025-09-26
fix: conditionally enable IPv6 listening

c3y1huang:

Conditionally enables IPv6 listening in the Longhorn UI Nginx container. This ensures the container works correctly on both IPv4-only and dual-stack networks.

### Issue

longhorn/longhorn#11869

### DIFF
--- a/scripts/start-nginx.sh
+++ b/scripts/start-nginx.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+set -e
+
+TEMPLATE=/etc/nginx/nginx.conf.template
+CONFIG_DIR=/var/config/nginx
+FINAL_CONF=$CONFIG_DIR/nginx.conf
+
+mkdir -p $CONFIG_DIR
+cp -r /etc/nginx/* $CONFIG_DIR
+
+# Detect IPv6 only
+if ip -6 addr show scope global | grep -q inet6; then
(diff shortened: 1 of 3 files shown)


derekbit (approved):
LGTM

derekbit:
@mergify backport v1.10.x

scripts/start-nginx.sh
```diff
@@ -0,0 +1,26 @@
...
+
+TEMPLATE=/etc/nginx/nginx.conf.template
+CONFIG_DIR=/var/config/nginx
+FINAL_CONF=$CONFIG_DIR/nginx.conf
+
+mkdir -p $CONFIG_DIR
+cp -r /etc/nginx/* $CONFIG_DIR
+
+# Detect IPv6 only
+if ip -6 addr show scope global | grep -q inet6; then
```
houhoucoop, line 13: [Nit] Do we need the flag (e.g. `ENABLE_IPV6`) to allow manual IPv6 enablement when auto detection fails?

c3y1huang: For now, I think no. Most distributions have the IPv6 kernel module enabled by default, even if IPv6 isn’t actively used. In such cases, Kubelet may still detect it and assign inet6 to pods, even when Kubernetes is running in a single IPv4 stack.

We can improve this later if needed.

PR #1005, opened 2026-02-02
fix: Update dashboard schedulable storage metrics

elTwingo:
This update uses storageScheduled to represent allocated (scheduled) capacity and storageAvailable to represent schedulable capacity, avoiding double counting and negative values caused by subtracting reserved space. The result is a more accurate and predictable view of how much storage is actually allocated by Longhorn versus what remains available for scheduling.

Issue :

https://github.com/longhorn/longhorn/issues/12633

### DIFF
--- a/src/routes/dashboard/components/resourceOverview.js
+++ b/src/routes/dashboard/components/resourceOverview.js
@@ -90,7 +90,7 @@ class ResourceOverview extends React.Component {
       const result = host.data.filter(n => n.allowScheduling === true).reduce((total, currentNode) => {
         return total + Object.values(currentNode.disks).filter(d => d.allowScheduling === true).reduce((availabeSpace, currentDisk) => {
           if (currentDisk.diskType === blockDiskType || !currentDisk.diskType) {
-            return availabeSpace + (currentDisk.storageAvailable - currentDisk.storageReserved)
+            return availabeSpace + currentDisk.storageAvailable
           }
           return availabeSpace
         }, 0)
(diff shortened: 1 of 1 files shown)


src/routes/dashboard/components/resourceOverview.js
```diff
@@ -103,7 +103,7 @@ class ResourceOverview extends React.Component {
       return host.data.filter(n => n.allowScheduling === true).reduce((total, currentNode) => {
         return total + Object.values(currentNode.disks).filter(d => d.allowScheduling === true).reduce((usedSpace, currentDisk) => {
           if (currentDisk.diskType === blockDiskType || !currentDisk.diskType) {
-            return usedSpace + (currentDisk.storageMaximum - currentDisk.storageAvailable)
+            return usedSpace + (currentDisk.storageMaximum - currentDisk.storageReserved - currentDisk.storageAvailable)
```
houhoucoop: Would using `storageScheduled` be more accurate here? also wrap the final result with `Math.max` to prevent negative values, e.g.
```js
const computeUsedSpace = (blockDiskType) => {
  const result = host.data
    .filter((n) => n.allowScheduling === true)
    .reduce((total, currentNode) => {
      return (
        total + Object.values(currentNode.disks)
          .filter((d) => d.allowScheduling === true)
          .reduce((used, currentDisk) => {
            if ( currentDisk.diskType === blockDiskType || !currentDisk.diskType ) {
              return used + (currentDisk.storageScheduled || 0);
            }
            return used;
          }, 0)
      );
    }, 0);

  return Math.max(0, result);
};
```

derekbit: @elTwingo Could you check the comment?

davidcheng0922: For my personal opinion, using `storageScheduled` seems more aligned with Longhorn semantics here. 

`storageAvailable` represents schedulable capacity, while `storageScheduled` reflects capacity already allocated by Longhorn. Using `storageMaximum - storageReserved - storageAvailable` avoids double-counting reserved space, but it still derives `Used` from `physical usage` rather than `Longhorn allocation`. 

I prefer @houhoucoop's suggestion here.

derekbit: cc @houhoucoop

houhoucoop:
Since this PR has been inactive for a while, I opened a separate PR to move this forward: https://github.com/longhorn/longhorn-ui/pull/1038.
CC @elTwingo @derekbit

PR #1032, opened 2026-04-14
feat: TooManySnapshots volume condition uses a hard-coded threshold despite configurable snapshot max count

houhoucoop:
- Use `min(snapshot-warning-threshold, v.Spec.SnapshotMaxCount)` as the snapshot count threshold

### Issue
[[UI][IMPROVEMENT] TooManySnapshots volume condition uses a hard-coded threshold despite configurable snapshot max count #12922](https://github.com/longhorn/longhorn/issues/12922)

### Test Result
https://github.com/user-attachments/assets/964005ac-a5b4-4225-8f1b-bf8ba713bff2

N/A

### DIFF
--- a/src/routes/volume/detail/ConditionTooltip.js
+++ b/src/routes/volume/detail/ConditionTooltip.js
@@ -9,10 +9,19 @@ const toolTipContent = (field, prefix, value = '') => {
   return (<div style={{ marginBottom: 5 }}>{prefix}: {text}</div>)
 }
 
-function ConditionTooltip({ selectedVolume, conditionKey }) {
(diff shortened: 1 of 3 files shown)


COLDTURNIP (approved):

houhoucoop:
@COLDTURNIP For the v2 data engine, should we still display `100`?

COLDTURNIP:
> @COLDTURNIP For the v2 data engine, should we still display `100`?

v2 data engine currently hasn't implement the hard limit of the amount of snapshot, but the warning condition logic is identical. We don't need to tell the data engine version here.

PR #1037, opened 2026-05-11
fix: UI does not update v2 backup state from Error to Completed

houhoucoop:
- Display `Successful` first when the state is `Completed`

### Issue
[[BUG] UI does not update v2 backup state from Error to Completed #12842](https://github.com/longhorn/longhorn/issues/12842)

### Test Result
Screenshot 2026-05-11 at 5 10 07 PM

N/A

### DIFF
--- a/src/utils/formatter.js
+++ b/src/utils/formatter.js
@@ -90,11 +90,12 @@ export function formatSnapshot(selectVolume, snapshot) {
         if (ele.error) {
           backupStatusErrorMsg.push({ replica: ele.replica, error: ele.error })
         }
-        total += ele.progress
+        total += ele.state === 'Completed' ? 100 : ele.progress
       })
       backupStatusObject = {}
       backupStatusObject.backupError = backupStatusErrorMsg
       backupStatusObject.progress = Math.floor(total / backupStatusObjectList.length)
+      backupStatusObject.isCompleted = backupStatusObjectList.every(ele => ele.state === 'Completed')
       backupStatusObject.snapshot = snapshot.name
       backupStatusObject.replicas = backupStatusObjectList.filter(item => item.replica).map(item => item.replica).join(', ')
       backupStatusObject.backupIds = backupStatusObjectList.filter(item => item.id).map(item => item.id).join(',')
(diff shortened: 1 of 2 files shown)


src/utils/formatter.js
```diff
@@ -90,11 +90,12 @@ export function formatSnapshot(selectVolume, snapshot) {
         if (ele.error) {
           backupStatusErrorMsg.push({ replica: ele.replica, error: ele.error })
         }
-        total += ele.progress
+        total += ele.state === 'Completed' ? 100 : ele.progress
```
derekbit, line 93: Won't the `ele.progress` be 100 if the state is completed? cc @mantissahz

mantissahz: When the backup sync monitor encountered an error, it would record the progress in `Backup.Status.Progress`and the error message in the `Backup.Status.Error`. And then after the backup is completed (Also in sync monitor, the `Backup.Status.State` is `Completed`), the error will not be cleaned up.
The progress and state are received from the engine.
And yes, it's weird when it's in `Completed` state and progress is not 100.
But it's actually completed.

It's another issue we should fix.

mantissahz: https://github.com/longhorn/longhorn-spdk-engine/blob/ed42dab52ca358980f02264d4693b0a3bca6c251/pkg/spdk/backup.go#L257-L263
It should be 100 already when it is set `ProgressStateComplete`.
The flow should be refactored in the backup controller reconciliation or the backup monitor synchronization

derekbit: Sounds a small fix. @mantissahz Would you be able to create a PR to fix this issue?

mantissahz: Sure, trying to figure out where the problem is.

mantissahz: https://github.com/longhorn/longhorn-manager/pull/4734

mantissahz:
Control plane fix:
https://github.com/longhorn/longhorn-manager/pull/4734

src/utils/formatter.js
houhoucoop: After the fix, I assume the progress will become 100, but will `ele.error` still have a value?

mantissahz: Yes, it depends on the backend.

houhoucoop:
Closed as won’t fix.

